PR #128, opened 2026-04-26
Remove UI_SPEC.md

MC-Meesh:
## Summary

- Deletes `UI_SPEC.md` — stale scaffold that was never kept in sync
- Drops the two `UI_SPEC §12.22` cross-references from `SPEC.md` and `ui/tests/e2e/app-logs-tab.spec.ts`
- `SPEC.md` is kept; it will be updated separately to reflect current state


MC-Meesh:
Folding UI_SPEC.md removal into #129 to keep things in one PR.

PR #255, opened 2026-05-08
fix: preview env lifecycle bugs (#223)

### LINKED ISSUE #223: Preview environment lifecycle bugs: short-circuit, secret leak, orphans, reopened PRs
## Summary

Five related bugs in `internal/controller/previewenvironment_controller.go` that cause preview environments to miss updates, leak secrets, persist as orphans, and fail to recreate on PR reopen.

## Details

**P0-9: Short-circuit skips spec change reconciliation (line 247-249)**

When the image tag contains the current SHA and phase is Ready, the controller returns `proceed=false`, skipping Deployment/Service/Ingress reconciliation. If spec fields change (replicas, resources, env vars, domain) without a new commit, the preview is never updated.

**P0-10: GC finalizer doesn't clean up Secrets (line 756-787)**

`gcPreviewResources` deletes Deployments, Services, and Ingresses but not Secrets. The preview namespace is kept alive intentionally, so namespace deletion won't cascade-clean these Secrets. Stale credentials accumulate over time.

**P0-11: "reopened" PR action not handled (webhook/handler.go:665-677)**

`normalizeAction` doesn't map `"reopened"` → `"opened"`. When a user reopens a closed PR, the preview (deleted on close) is never recreated. Users must push a new commit as a workaround.

**P2-3: Missing owner reference from PE to parent App**

No `SetControllerReference` or `SetOwnerReference` anywhere in the PE controller. If an App is deleted while a preview is active, the PE persists as an orphan with no garbage collection.

**P2-13: Preview Secret labels missing pr-number (line 428-433)**

Labels on env Secrets don't include `mortise.dev/pr-number`, which the GC selector uses. Even if Secrets are added to GC, they won't match.

## Fix

1. Add spec-hash comparison so spec changes trigger reconciliation even without new SHA
2. Add Secret deletion to `gcPreviewResources` and add `mortise.dev/pr-number` to Secret labels
3. Map `"reopened"` → `"opened"` in `normalizeAction`
4. Add `controllerutil.SetControllerReference(app, pe, r.Scheme)` when creating/updating PE

MC-Meesh:
## Summary
- Add owner references on PreviewEnvironments so App deletion garbage-collects orphan PEs
- Fix short-circuit logic: return `true` so spec changes still reconcile when image SHA matches
- Clean up secrets in `gcPreviewResources` to prevent secret leaks on PE deletion
- Handle `reopened` PR events in webhook handler alongside `opened`

Closes #223

## Test plan
- [ ] Verify PE gets owner reference set on reconcile
- [ ] Confirm spec changes apply even when image SHA hasn't changed
- [ ] Check secrets are cleaned up when PE is garbage collected
- [ ] Test reopened PR triggers preview environment creation


MC-Meesh (commented):
Code quality audit: no significant issues found.

Changes reviewed:
- Owner reference for PE garbage collection: correct use of `controllerutil.SetControllerReference`
- Short-circuit fix (`false` -> `true`): correctly allows Deployment/Service/Ingress reconciliation to proceed even when the image SHA hasn't changed, so spec changes (replicas, env vars, etc.) are picked up
- GC error handling: good improvement — previously all Delete errors were silently discarded, now they're collected and returned as a joined error
- Secret cleanup added to GC: good catch, was previously missing
- GitLab `reopened` event: correct fix
- PR-number label on preview env secret: useful for debugging

All changes are well-scoped and correct.

MC-Meesh (commented):

**Findings:**

### WARNING
- **`gcPreviewResources` does not filter `IsNotFound` errors on Delete** (`internal/controller/previewenvironment_controller.go`): The GC loop now properly collects errors instead of discarding them, which is good. However, during concurrent reconciliation or when a namespace is being torn down, a resource may be deleted between the List and Delete calls. The current code treats `IsNotFound` as a real failure, which will cause spurious error returns and potentially block PE finalization. Each delete should wrap with `if client.IgnoreNotFound(err) != nil` (the standard controller-runtime pattern) rather than treating every delete error as fatal.

- **Owner reference set before validation** (`internal/controller/previewenvironment_controller.go`, around the `hasOwnerRef` block): The owner reference is stamped and persisted via `r.Update` before the "is this a git source app?" check. If the PE references a non-git-source app, the owner ref is still written (wasted API call + unnecessary GC coupling). Consider moving the owner-ref block after the source-type guard, or at least after the app lookup succeeds. Not a correctness bug since SetControllerReference on a non-git app is harmless, but it's a wasted write on an invalid PE.

### NOTE
- **Secrets now GC'd in preview teardown** — the addition of Secret cleanup in `gcPreviewResources` is a good catch. Verify that the label selector used (`previewLabels`) is applied to Secrets created by `reconcilePreviewEnvSecret`. The new `mortise.dev/pr-number` label added to the Secret in this same PR should ensure it matches.

- **GitLab `reopened` MR events now handled** (`internal/webhook/handler.go`): Clean single-line fix. Correct behavior — a reopened MR should trigger the same flow as an opened one.

- **Build short-circuit now returns `true` for proceed** — this is the core behavioral fix. Previously, updating replicas/resources/env on an already-built PE had no effect until the next SHA change. The comment update accurately describes the new behavior.

MC-Meesh:
Audit follow-up on the latest head (`b9d7a95`):

- The two prior warnings are resolved in the current patch: owner reference setup now happens after the git-source guard, and preview GC ignores `IsNotFound` delete races for Deployments, Services, Ingresses, and Secrets.
- Confirmed the Secret label selector path: `reconcilePreviewEnvSecret` now applies the same `previewLabels` set used by `gcPreviewResources`, so preview Secrets are included in teardown.
- No new behavioral regressions found in the updated diff.

Verification:
- `go test ./internal/webhook` passed.
- `KUBEBUILDER_ASSETS="$(./bin/setup-envtest use 1.35 --bin-dir "$(pwd)/bin" -p path)" go test ./internal/controller` passed.

Remaining status: GitHub Actions CI for this head is still queued, and this PR still has no changed test files despite covering lifecycle behavior.

PR #257, opened 2026-05-08
fix: API handler cleanup — validation, error handling, consistency (#241)

### LINKED ISSUE #241: API handler cleanup: validation, error handling, consistency
## Summary

Several API handler issues that can be fixed in a single cleanup PR.

## Details

**P2-1: UpdateApp replaces entire spec without validation (apps.go:214-250)**
\`UpdateApp\` doesn't validate environment names via \`validateDNSLabel\` unlike \`CreateApp\`. Invalid DNS names pass through and cause controller failures.

**P2-7: authorize-before-resolve pattern (pods.go:50-54, traffic.go:39-43,108-112)**
These handlers call \`s.authorize()\` with the raw URL param before \`s.resolveProject()\` validates it. Not exploitable but inconsistent with other handlers.

**P2-8: time.Now() in API handlers**
Direct \`time.Now()\` calls in: \`members.go:183\`, \`rebuild.go:239\`, \`env.go:496\`, \`traffic.go:135\`. Should use injected clock for testability.

**P2-9: Inconsistent token hash key names (tokens.go:110 vs 374)**
Per-app tokens use \`token-hash\` (hyphen), project tokens use \`token_hash\` (underscore). Works but maintenance hazard.

**P3-4: writeError leaks k8s internal details (apps.go:291-305)**
The 500 catch-all passes \`err.Error()\` directly into JSON responses. K8s API errors contain internal cluster details.

**P3-5: CreateSecret doesn't validate name as DNS label (secrets.go:65-107)**
Non-empty check only. Invalid names produce raw k8s errors instead of clean validation messages.

## Fix

1. Add \`validateDNSLabel\` to \`UpdateApp\`
2. Reorder to resolve-then-authorize in pods/traffic handlers
3. Inject \`clock.Clock\` into API server struct
4. Standardize token hash key naming
5. Return generic message for 500s, log details server-side
6. Add \`validateDNSLabel\` to \`CreateSecret\`

MC-Meesh:
## Summary
- Add DNS label validation on environment names in UpdateApp
- Inject clock via `s.clock()` for testability (replaces direct `time.Now()` calls)
- Stop leaking internal error messages to clients (`writeError` now returns generic message + logs detail)
- Fix pods handler: resolve project namespace before authz check (was authorizing against empty namespace)

Closes #241

## Test plan
- [ ] Submit invalid environment name — should get 400 with validation message
- [ ] Verify internal errors return generic "internal server error" (not stack traces)
- [ ] Confirm pod listing works with correct namespace-scoped authorization


MC-Meesh (commented):
**[WARNING]** internal/api/apps.go:writeError — The `writeError` function now correctly stops leaking internal error details to clients (good security fix), but the `slog.Error` call logs without any request context (no request ID, path, method). In a high-traffic API, correlating these generic "internal API error" log lines back to specific requests will be difficult. Consider threading `*http.Request` or at minimum the request path into the log call.

**[NOTE]** internal/api/pods.go + internal/api/traffic.go — The authorization reorder (resolve project first, then authorize with namespace) is a meaningful security fix. The old code was calling `authorize` with an empty `Namespace` field, which depending on the authz implementation could have been either a no-op or overly permissive. Good catch.

All other changes (injectable clock for testability, DNS label validation on UpdateApp, secret name validation) are clean and correct.

MC-Meesh (commented):

**Findings:**

### NOTE
- **`writeError` signature change is a cross-PR merge hazard**: This PR changes `writeError(w, err)` to `writeError(w, r, err)` across all API handlers. PR #259 (authz fixes) adds new `writeError(w, err)` calls in `RedeployStale` (`internal/api/rebuild.go`). Whichever PR merges second will have a compilation failure. Not a bug in this PR, but coordinate merge order — #257 should merge first, then #259 needs to update its `writeError` calls to include the `r` parameter.

- **`pokeAppForReconcile` now takes explicit `time.Time`** — good fix, eliminates a `time.Now()` call in API handler code. Uses `s.clock().Now()` at the call site, which is the correct pattern for testability.

- **Internal errors no longer leak to clients** — the `writeError` change from `errorResponse{err.Error()}` to `errorResponse{"internal server error"}` with server-side logging is a solid security improvement. The `slog.Error` includes method, path, and request ID for debugging.

- **Environment name validation on UpdateApp** — the new `validateDNSLabel` loop in `UpdateApp` closes a gap where CreateApp validated env names but UpdateApp did not. Good catch.

- **`time.Now()` removed from `AddMember`** — replaced with `s.clock().Now()`, consistent with the project convention.

- **ListPods: authorize after resolveProject** (`internal/api/pods.go`) — the reordering is correct. The old code called `chi.URLParam(r, "project")` directly and used it for authz before resolving the project (which validates it exists). The new order resolves first, then authorizes with the validated namespace.

No issues found that would block merge. Clean to merge.

PR #258, opened 2026-05-08
fix: observer goroutine leaks, missing sweep, unbounded latency (#224)

### LINKED ISSUE #224: Observer: goroutine leaks, missing sweep, unbounded latency growth
## Summary

Multiple resource management bugs in `cmd/observer/` cause goroutine leaks, silent log loss, unbounded memory growth, and stale cache entries.

## Details

**P0-12: Tailer goroutine leak on stream failure (log_collector.go:172-217)**

When `tailPod` exits on stream error/EOF, the entry in `c.tailers[key]` is never removed. The stale entry prevents a new tailer from starting (line 152: `if _, exists`), silently losing logs for running pods. Same pattern in `traffic_collector.go:180-209`.

**P0-13: Secondary goroutine leak on context cancellation (log_collector.go:189-192)**

A secondary goroutine waits on `<-ctx.Done()` to close the stream. If the stream closes naturally before context cancellation, this goroutine hangs forever. Same in `traffic_collector.go:196-199`.

**P1-11: LiveTrafficCache.Sweep() never called (live_traffic_cache.go:135)**

`LiveMetricsCache.Sweep()` is called in `metrics_collector.go:104`, but `LiveTrafficCache.Sweep()` is never called. Dead keys accumulate indefinitely.

**P2-5: Unbounded latency slice growth (traffic_collector.go:39,256)**

`trafficBucket.latencies []float64` grows by one entry per request until flush. Under 10K req/s with a 5-second bucket, that's 50K+ entries. If flush stalls, accumulation is unbounded.

## Fix

1. Remove tailer map entry on goroutine exit so `sync()` can restart it
2. Use `select` on both `<-ctx.Done()` and a `done` channel for the stream-closer goroutine
3. Add `liveTrafficCache.Sweep()` call alongside the metrics sweep
4. Cap the latency slice (reservoir sampling for percentile accuracy with fixed memory)

MC-Meesh:
## Summary
- Fix goroutine leaks in observer by ensuring proper context cancellation and cleanup
- Add missing sweep logic to prevent unbounded latency growth
- Ensure collector goroutines are properly bounded and terminated

Closes #224

## Test plan
- [ ] Verify observer goroutines are cleaned up on shutdown
- [ ] Confirm sweep runs and prevents stale data accumulation
- [ ] Check latency metrics remain bounded under sustained load


MC-Meesh (commented):
**[WARNING]** cmd/observer/traffic_collector.go:processLine — The reservoir sampling uses `c.rng.IntN(int(b.latencyCount))` where `b.latencyCount` is `int64` but is cast to `int`. On a 64-bit system this is fine, but if `latencyCount` ever overflows `int` (>2^63 on 64-bit, >2^31 on 32-bit), `IntN` will panic with a negative argument. This is unlikely in practice since flush() resets buckets periodically, but a defensive `int(b.latencyCount % (maxLatencySamples * 1024))` or similar cap would prevent the edge case. Low risk given flush cadence.

**[NOTE]** cmd/observer/log_collector.go:startTailer — The cleanup `defer` checks `if _, exists := c.tailers[key]` before decrementing `tailerCount`. This is correct — it guards against double-decrement if the tailer was already removed by `syncTailers`. Good defensive pattern.

All other changes are clean:
- The `done` channel pattern in both `tailPod` and `tailTraefikPod` correctly prevents goroutine leaks when the stream ends naturally (EOF) rather than via context cancellation
- `c.rng` access is protected by `accMu` — thread-safe
- `liveCache.Sweep()` placement in `flush()` is correct
- Tests cover the key scenarios adequately

MC-Meesh (commented):

**Findings:**

### NOTE
- **Reservoir sampling `latencyCount` overflow guard is safe but worth a comment**: `latencyCount` is `int64` and the guard `if n > math.MaxInt` protects the `IntN(int(n))` call on 32-bit platforms where `int` is 32 bits. On 64-bit (which is the realistic deployment target), `math.MaxInt == math.MaxInt64` so the guard is a no-op. This is fine — it's defensive — but a brief comment explaining the 32-bit motivation would help future readers.

- **`rng` thread safety is correct**: `c.rng.IntN()` is called inside the `c.accMu.Lock()` scope within `processLine`, so no data race.

- **Tailer cleanup defer pattern is correct**: Both `LogCollector.startTailer` and `TrafficCollector.tailTraefikPod` now have deferred cleanup that removes the map entry and decrements the counter under the mutex. The `done` channel pattern for the stream-closing goroutine prevents the goroutine leak where the closer goroutine would outlive the tailer.

- **`liveCache.Sweep()` in `flush()`** — this was the missing sweep call from the original issue. Positioned correctly after accumulator drain and before store insert.

- **Test coverage is solid**: Tests verify tailer cleanup on context cancellation for both log and traffic collectors, and the reservoir sampling cap test verifies the `maxLatencySamples` bound.

No issues found — clean to merge.

MC-Meesh:
Audit follow-up for mo-e8k:

- Re-reviewed the current diff in `cmd/observer/log_collector.go`, `cmd/observer/traffic_collector.go`, and `cmd/observer/collectors_test.go`.
- No additional actionable regressions found. Tailer cleanup, closer goroutine shutdown, bounded latency reservoir sampling, and `LiveTrafficCache.Sweep()` placement look correct.
- Local verification: `go test ./cmd/observer` passes.
- Current GitHub checks: Go vet/staticcheck, Helm lint/template tests, Unit + envtest, and UI typecheck/build pass; Integration tests are still pending.
- PR is still blocked on required review / pending integration status, not on a code issue found in this audit.

PR #259, opened 2026-05-08
fix: close viewer project-bypass and enforce restricted-env authz (#231)

### LINKED ISSUE #231: Authz bypass: viewer reads all projects, restricted envs unenforced
## Summary

Two authorization bugs that together mean: (1) platform viewers can read all project data without membership, and (2) the restricted-environment feature is completely non-functional.

## Details

**P1-1: Platform viewer bypasses project membership (authz/native.go:29-31)**

```go
if p.Role == auth.RoleViewer {
    return action == ActionRead
}
```

This early return skips `authorizeProject()` entirely. A platform viewer can read any project-scoped resource (apps, secrets, tokens, members, env vars) across all projects without membership.

**P1-2: Restricted environment authz dead code (14+ handlers)**

The `authz.Resource` struct has an `Environment` field for restricted-env checks, but **no API handler ever populates it**. The `authorizeDeveloper` restricted-env check is dead code.

Affected handlers:
- `exec.go:55`, `rebuild.go:110,172`, `rollback.go:46,141`, `deploy.go:64`
- `env.go:99,183,282`, `build_args.go:69`, `secrets.go:70,169`, `domains.go:92,180`

A developer-role user can deploy, exec, modify env vars, and manage secrets in restricted (e.g., production) environments.

## Fix

1. Remove the viewer early return; let viewers fall through to project membership checks
2. Add `Resource.Environment` to every handler that operates on a specific environment

MC-Meesh:
## Summary
- Close authorization bypass where viewers could read all projects regardless of membership
- Enforce restricted environment access controls that were previously unenforced
- Add authz tests for viewer project membership and restricted env checks

Closes #231

## Test plan
- [ ] Verify viewer can only access projects they are members of
- [ ] Confirm restricted environments are inaccessible to unauthorized roles
- [ ] Run `make test` — all authz tests pass

### DIFF
--- a/internal/api/build_args.go
+++ b/internal/api/build_args.go
@@ -66,7 +66,7 @@ func (s *Server) GetBuildArgs(w http.ResponseWriter, r *http.Request) {
 // @Router /projects/{project}/apps/{app}/build-args [put]
 func (s *Server) PutBuildArgs(w http.ResponseWriter, r *http.Request) {
 	projectName := chi.URLParam(r, "project")
-	if !s.authorize(w, r, authz.Resource{Kind: "app", Project: projectName}, authz.ActionUpdate) {
+	if !s.authorize(w, r, authz.Resource{Kind: "app", Project: projectName, Environment: queryEnv(r)}, authz.ActionUpdate) {
 		return
 	}
 	app, envName, ok := s.resolveAppEnv(w, r)
--- a/internal/api/deploy.go
+++ b/internal/api/deploy.go
@@ -61,7 +61,7 @@ func (s *Server) Deploy(w http.ResponseWriter, r *http.Request) {
 
 	// Check auth: JWT principal (policy-checked) or deploy token (inline check).
 	if p := PrincipalFromContext(r.Context()); p != nil {
-		if !s.authorize(w, r, authz.Resource{Kind: "app", Namespace: ns, Project: projectName}, authz.ActionUpdate) {
+		if !s.authorize(w, r, authz.Resource{Kind: "app", Namespace: ns, Project: projectName, Environment: req.Environment}, authz.ActionUpdate) {
 			return
 		}
 	} else {
--- a/internal/api/domains.go
+++ b/internal/api/domains.go
@@ -89,7 +89,7 @@ func (s *Server) ListDomains(w http.ResponseWriter, r *http.Request) {
 // @Router /projects/{project}/apps/{app}/domains [post]
 func (s *Server) AddDomain(w http.ResponseWriter, r *http.Request) {
 	projectName := chi.URLParam(r, "project")
-	if !s.authorize(w, r, authz.Resource{Kind: "app", Project: projectName}, authz.ActionUpdate) {
+	if !s.authorize(w, r, authz.Resource{Kind: "app", Project: projectName, Environment: queryEnv(r)}, authz.ActionUpdate) {
 		return
 	}
 	app, envName, ok := s.resolveAppEnv(w, r)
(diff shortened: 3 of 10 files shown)


MC-Meesh (commented):
**Code Quality Audit — PR #259**

**[WARNING]** internal/api/rollback.go:Promote — The Promote handler only checks authz against `req.To` (destination environment), not `req.From` (source environment). A developer restricted from a production environment could still call Promote with `from: "production"` to read production's current image and deploy state. Consider adding a `ActionRead` check on `req.From` as well:
```go
if !s.authorize(w, r, authz.Resource{Kind: "app", Namespace: ns, Project: projectName, Environment: req.From}, authz.ActionRead) {
    return
}
```

**[WARNING]** internal/api/rebuild.go:RedeployStale — If `s.authz.Authorize()` returns an error midway through the environment loop (after some environments have already been restarted via `restartDeployment`), the handler returns an HTTP error but the already-restarted deployments are not reflected in the response or status update. The caller has no way to know which environments were actually restarted. Consider collecting errors and returning a partial-success response instead of short-circuiting.

**[NOTE]** internal/api/rebuild.go:RedeployStale — The original handler had a project-level authz check before entering the loop. The new per-environment check removes that gate, so an unauthenticated request (where `PrincipalFromContext` returns nil) is now caught explicitly — good. However, the new `PrincipalFromContext` nil-check duplicates what `s.authorize` would typically handle. If there is upstream middleware guaranteeing auth, this is fine; otherwise, confirm the handler is only reachable via authenticated routes.

MC-Meesh (commented):

**Findings:**

### CRITICAL
- **`writeError(w, err)` calls in `RedeployStale` will not compile after PR #257 merges** (`internal/api/rebuild.go`, lines ~187 and ~192 in the new code): PR #257 changes `writeError` from `writeError(w, err)` to `writeError(w, r, err)`. This PR adds two new `writeError(w, err)` calls inside `RedeployStale`. Whichever PR merges second will fail to compile. These need to be `writeError(w, r, err)`.

### WARNING
- **`Promote` only checks authz on `req.To`, not `req.From`** (`internal/api/rollback.go`): The authorization check uses `Environment: req.To` (the target environment). A developer with access to staging but not production could promote FROM production TO staging, reading production's deployment state in the process. Consider whether read access on the source environment should also be enforced, or whether the current design intentionally allows this (if so, document why).

### NOTE
- **Viewer bypass fix is correct** (`internal/authz/native.go`): Previously, `auth.RoleViewer` was granted blanket read access to all project-scoped resources regardless of project membership. The fix gates project-scoped viewer access through `authorizeProject`, which checks for a ProjectMember CRD. This is the right fix.

- **Restricted-env enforcement is comprehensive**: All env-mutating endpoints (PutEnv, PatchEnv, ImportEnv, Deploy, Redeploy, Rollback, Promote, ExecInApp, CreateSecret, DeleteSecret, PutBuildArgs, AddDomain, ListDomains) now pass the environment to the authz resource. This means `NativePolicyEngine.authorizeProject` can enforce the `Restricted` flag on project environments.

- **`RedeployStale` per-env authz with skip list** — the approach of skipping unauthorized environments (rather than rejecting the whole request) is a good UX choice for a bulk operation. The `skipped` field in the response makes the behavior transparent to the caller.

- **Test coverage is thorough**: Three new test cases cover viewer-without-membership denial, viewer-with-membership read-only, and developer restricted-env enforcement. These directly validate the security invariants.

MC-Meesh:
Audit follow-up finding from mo-7qp:

**CRITICAL:** `internal/authz/native.go` now lets a platform `viewer` inherit write privileges from a higher project membership role. The new viewer branch delegates project-scoped resources directly to `authorizeProject`, and `authorizeProject` applies the `ProjectMember` role without capping it by the platform role. That means a principal with `auth.RoleViewer` plus a `ProjectMember` of `developer` or `owner` can pass `ActionUpdate`/other write checks for that project, despite the platform role contract in `SPEC.md` saying viewer is read-only.

I verified this against the current PR head by temporarily adding this regression test; it fails:

```go
viewer := auth.Principal{ID: email, Email: email, Role: auth.RoleViewer}
ok, err := engine.Authorize(ctx, viewer, Resource{Kind: "app", Project: project}, ActionUpdate)
// ok is true when ProjectMember role is developer
```

Suggested fix: in the `auth.RoleViewer` branch, keep the project membership check for project-scoped reads, but hard-deny non-read actions before delegating to project RBAC. Add a regression test for a platform viewer with `ProjectRoleDeveloper` or `ProjectRoleOwner` membership attempting `ActionUpdate`.

PR #262, opened 2026-05-08
fix: strengthen exec handler label selector and output limit (#236)

### LINKED ISSUE #236: Exec handler: weak label selector + no output size limit
## Summary

The exec handler uses a weaker pod label selector than other handlers, and has no restrictions on command output size.

## Details

**P1-7: Weak label selector (exec.go:99-101)**

`findAppPod` uses only `app.kubernetes.io/name={appName}`. The log handler uses three labels: `AppNameLabel`, `managed-by: mortise`, `mortise.dev/environment`. The exec handler could match non-Mortise pods or pods in the wrong environment.

**P2-2: No output size limit (exec.go:50-95)**

The handler accepts arbitrary commands with unbounded `bytes.Buffer` for stdout/stderr. A `yes` command could OOM the operator. No audit logging of exec commands.

## Discussion needed

How restrictive should exec be?
- **Output limit** — add a max output size (e.g., 10 MiB)?
- **Audit logging** — log all exec commands for compliance?
- **Command restrictions** — is an allow-list appropriate, or is exec inherently a power-user feature that should just be auth-gated?

MC-Meesh:
## Summary
- Tighten exec handler's pod label selector to prevent cross-tenant access
- Add output size limit to prevent unbounded memory usage from exec streams
- Add LimitedBuffer with tests for safe output capping

Closes #236

## Test plan
- [ ] LimitedBuffer unit tests pass
- [ ] Exec handler correctly limits output
- [ ] CI pipeline green


MC-Meesh (commented):
Code quality audit: no significant issues found.

The `limitedBuffer` implementation is correct — it reports all bytes as consumed to avoid upstream write errors while silently capping at the limit. Truncation detection with `>=` is accurate. The strengthened label selector properly scopes pod selection to mortise-managed pods in the correct environment, closing a real security gap where exec could hit unrelated pods. Tests have good coverage of boundary conditions.

MC-Meesh (commented):

**Findings:**

### NOTE
- **`limitedBuffer.Write` silently succeeds after truncation — no signal to caller**: When the buffer is full, `Write` returns `len(p), nil` — the `io.Writer` contract says the return value is the number of bytes written, and returning more than actually stored is technically a violation. In practice this works because `remotecommand.StreamOptions` does not check for short writes, but it could cause subtle bugs if `limitedBuffer` is reused elsewhere. Consider returning `(len(p), nil)` only when all bytes fit, and `(written, io.ErrShortWrite)` or a custom error otherwise — or at minimum add a comment explaining the intentional contract violation and why it is safe for the exec use case.
- **`truncated` detection is at-limit, not over-limit**: `truncated` is set when `stdout.Len() >= maxExecOutputBytes`. Since the buffer stores exactly `limit` bytes and discards the rest, `Len() == limit` means data *may* have been truncated (the output was exactly 10 MiB) or it was coincidentally exactly 10 MiB. This is a minor false-positive edge case — not a real problem, but worth a comment.
- **No integration or E2E test for the strengthened label selector**: The label selector change in `findAppPod` (adding `managed-by=mortise` and environment label) is the security-critical part of this PR, but it has no test coverage. The `limitedBuffer` unit tests are good, but the actual pod-selection hardening is untested. An envtest or integration test that creates pods with and without the expected labels and verifies the correct pod is selected would give confidence this works as intended.

PR #263, opened 2026-05-08
feat: user-friendly error for project name collisions (#183)

### LINKED ISSUE #183: Project name collisions are opaque to non-admin members
## Problem

Projects are cluster-scoped CRDs, so names must be globally unique across the entire platform. However, non-admin members can only see projects they belong to (via `ProjectMember`). This creates a blind collision:

1. User A creates project `api-gateway`
2. User B (not a member of that project) tries to create `api-gateway`
3. User B gets HTTP 409 with the raw Kubernetes error: `projects.mortise.dev "api-gateway" already exists`
4. User B has no idea why — they can't see the conflicting project, and the error gives no guidance

This is confusing UX. The user has no way to discover the conflict or resolve it without escalating to an admin.

## Current behavior

**API layer** (`internal/api/projects.go`): `CreateProject` calls `s.client.Create()`, and on failure, `writeError()` (`internal/api/apps.go:291`) passes through the raw k8s `AlreadyExists` error as a 409:

```go
if errors.IsAlreadyExists(err) {
    writeJSON(w, http.StatusConflict, errorResponse{err.Error()})
}
```

**Controller layer** (`internal/controller/project_controller.go`): Has a `checkControlNamespaceUniqueness` function that catches namespace-level collisions, but by that point the CRD already exists — the API-level collision happens first.

**Visibility layer** (`internal/api/projects.go` `ListProjects`): Non-admin users only see projects where they have a `ProjectMember`. The conflicting project is invisible to them.

### 1. User-friendly error message (low effort, high value)

In `CreateProject`, detect `IsAlreadyExists` specifically and return a clear message instead of the raw k8s error:

```
A project named "api-gateway" already exists. Project names must be unique
across the platform. If you need access to the existing project, ask a
project owner or platform admin to add you as a member.
```

This doesn't leak information about who owns the project — it just explains the constraint and tells the user what to do.

### 2. UI guidance

The create-project form should surface this error inline with the helpful message above, not a generic "conflict" toast.

### 3. Consider name availability check (optional)

Add a lightweight `GET /v1/projects/{name}/available` endpoint (or similar) that returns whether a project name is taken, so the UI can check on blur/type and show "name unavailable" before the user submits. This is safe — knowing a name is taken doesn't leak project details.

### 4. Longer-term: scoped project names? (discussion)

The current design ties project names 1:1 to Kubernetes namespaces (`pj-{name}`), which forces global uniqueness. For small teams this is fine, but at scale with many isolated teams, global uniqueness becomes friction. Options to discuss:

- **Accept it**: Railway and Vercel have globally unique project names. It's a known tradeoff.
- **Slug disambiguation**: Allow display names to collide but generate unique slugs (`api-gateway`, `api-gateway-2`). Namespace mapping uses the slug.
- **Org/team scoping**: Prefix with org/team (`teamA-api-gateway`). Requires an org concept that doesn't exist yet.

Given Mortise's current scope (small teams, single cluster), option 1 with a good error message (items 1-3) is probably sufficient for v1.

## Affected code

- `internal/api/projects.go` — `CreateProject` handler
- `internal/api/apps.go:291` — `writeError` generic handler
- `internal/controller/project_controller.go:361` — `checkControlNamespaceUniqueness`
- `internal/authz/native.go` — project-scoped auth via `ProjectMember`
- `api/v1alpha1/project_types.go:159` — cluster-scoped CRD definition

MC-Meesh:
## Summary
- Add server-side validation for duplicate project names with clear error message
- Surface collision errors in the UI new-project form
- Add unit tests for collision detection

Closes #183

## Test plan
- [ ] Creating a project with a duplicate name shows a user-friendly error
- [ ] Unit tests for name collision validation pass
- [ ] CI pipeline green


MC-Meesh (commented):
**[NOTE]** ui/src/routes/projects/new/+page.svelte — The `checkNameAvailability` function fires on blur with no debounce or throttle. If the field is focused/blurred rapidly (e.g., tabbing through a form), it will fire multiple concurrent requests. Consider debouncing (300-500ms) or cancelling the previous in-flight request to avoid a stale response overwriting a fresh one (race condition where an older response arrives after a newer one, showing incorrect status).

**[NOTE]** internal/api/projects.go:CheckProjectNameAvailable — This endpoint allows any authenticated user to enumerate all project names across the platform. If project names are considered sensitive (e.g., they reveal unreleased product names or client names), consider scoping visibility or adding rate limiting. If project names are intentionally public, this is fine as-is.

MC-Meesh (commented):

**Findings:**

### WARNING
- **UI help text still says `project-{name}` instead of `pj-{name}`**: The namespace hint in the new project form reads `Apps run in namespace project-{name}`. Per the architecture rules in CLAUDE.md, the current convention is `pj-{name}` for control namespace. This was carried over from the pre-existing code (not introduced by this PR), but since this PR is touching exactly this line, it should be corrected.

### NOTE
- **`CheckProjectNameAvailable` reveals project existence to any authenticated user**: The endpoint returns `available: false` for existing projects. Since any authenticated user can call it (per the Swagger annotation), this allows enumeration of all project names on the platform. This may be acceptable given that project names are not secret (they appear in DNS), but it is worth an explicit decision. If project names should not be enumerable, this endpoint should be restricted or rate-limited.
- **Route ordering relies on chi's longest-match semantics**: `/projects/{name}/available` is registered before `/projects/{project}`. This works because chi matches literal path segments before wildcards. However, a project literally named `available` would be unreachable via `GET /projects/available` (it would match the availability check route pattern instead). Since `available` is a valid DNS label, consider either documenting this as a reserved name or using a different URL pattern like `/projects/check-name/{name}`.
- **Debounce timer not cleaned up on component unmount**: `debounceTimer` from `setTimeout` is not cleared when the Svelte component is destroyed. If the user navigates away quickly, the stale callback may fire and attempt to update `nameStatus` on an unmounted component. In Svelte 5 with runes this likely just silently updates a dead reactive variable, but an `onDestroy` cleanup would be more correct.

MC-Meesh:
Audit follow-up on head `799a3bb68c4fd58f2b01e3256bb8ef2e2b9f7599`:

- No blocking behavioral regression found in the project collision handling diff.
- The earlier namespace help-text warning is resolved: the form now shows `pj-{name}`.
- The stale-response race is addressed with debouncing plus a sequence guard. Minor cleanup remains possible: clear `debounceTimer` on component destroy, but I do not consider it merge-blocking.
- The route-ordering concern around a project named `available` does not appear valid: `GET /projects/available` is a two-segment project route, while the availability route is `GET /projects/{name}/available` with an extra segment.
- The availability endpoint still intentionally exposes name existence to any authenticated user. This is already partly exposed through duplicate project creation attempts, but it is still a product/security decision worth accepting explicitly if project names are considered sensitive.

Verification:
- GitHub CI currently has Unit + envtest, Go vet + staticcheck, Helm lint/template, and UI typecheck/build passing.
- Integration tests were still running/pending when checked.
- Local `go test ./internal/api -run 'Test(CreateProjectDuplicateNameFriendlyError|CheckProjectNameAvailable)'` could not run because `/usr/local/kubebuilder/bin/etcd` is missing.
- Local `npm --prefix ui run check` could not run because `ui/node_modules` is absent (`svelte-kit` not found).

PR #264, opened 2026-05-08
fix: deploy handler sets per-environment image instead of global (#221)

### LINKED ISSUE #221: Deploy handler ignores environment — patches global image affecting all envs
## Summary

The Deploy handler patches `app.Spec.Source.Image` globally (deploy.go:105), ignoring the `environment` field in the request body. Every environment picks up the new image. A staging deploy token can inadvertently update production.

## Details

**deploy.go:105** — `app.Spec.Source.Image = req.Image` (global spec-level image)

The `environment` field is only used for activity logging (lines 112-114) and deploy token scope validation (line 90). The actual image patch is global.

**deploy.go:67-96** — When a deploy token is used (no JWT principal), the handler does not call `s.authorize()` at all. Even if restricted-env authz is fixed (see #N), the deploy token path bypasses it entirely.

**env.go:280-333** — `ImportEnv` may also replace instead of merge (the comment says "merge" but behavior depends on `store.Merge` implementation).

## Impact

- A deploy-token scoped to "staging" changes the image for production
- Users expect per-environment deploys but get global ones
- Project-scoped deploy tokens skip all environment validation

## Discussion needed

How should per-environment deploys work?

Options:
1. Store deployed image per-environment in `app.Spec.Environments[env].Image`
2. Store in status per-environment (e.g., `status.Environments[env].DeployedImage`)
3. Use the existing `LastBuiltImage` in per-env status

This affects the data model and how the controller resolves which image to deploy per env.

MC-Meesh:
## Summary
- Fix deploy handler to set container image per-environment rather than globally on the App spec
- Add CRD field for per-environment image override
- Update app controller to respect per-env image during reconciliation

Closes #221

## Test plan
- [ ] Deploying to one environment doesn't change the image for other environments
- [ ] Server test validates per-env image routing
- [ ] CI pipeline green


MC-Meesh (commented):
**[WARNING]** internal/api/deploy.go — There is no mechanism to clear a per-environment image override once set. If a user deploys `nginx:1.26.0` to staging via the environment-scoped path, then later wants staging to fall back to the spec-level image, they would need to explicitly deploy the same image as the spec-level one. Consider supporting an empty string or null to clear the override, or document this behavior so operators are not surprised.

**[NOTE]** internal/api/server_test.go:TestDeployPerEnvironment — The test verifies that the staging image was set and the spec-level image was untouched, but does not assert that the production environment's `Image` field remains empty. Adding `if productionImage != "" { t.Error(...) }` would strengthen the test against accidental cross-environment mutation.

MC-Meesh (commented):

**Findings:**

### WARNING
- **CRD schema change without migration path.** Adding `Environment.Image` is additive and optional, so existing CRDs won't break. However, there is no mechanism to clear a per-env image override once set. If a user deploys `nginx:1.26.0` to staging via the per-env path and later wants staging to track the spec-level image again, there is no API to reset `env.Image` back to empty. Consider whether the deploy handler (or a separate endpoint) should support clearing the override, or document that re-deploying with the spec-level image value is the workaround.

- **Duplicate image resolution logic.** The image precedence logic (per-env built > per-env spec > global spec) is implemented in two places: once inline in `Reconcile()` (line ~221) and once in `currentImageForEnv()` (line ~731). If the precedence changes, both must be updated in lockstep. The inline block in `Reconcile()` should call `currentImageForEnv()` instead of reimplementing the fallback chain.

### NOTE
- **Test creates app without asserting creation success.** In `TestDeployPerEnvironment`, the initial `doRequest` for app creation ignores the response code. If that request fails, the subsequent deploy test will fail with a confusing error. Adding `if w.Code != http.StatusCreated` after the create call would make test failures more diagnosable.

- **No test for the "environment not found" 404 path.** The handler correctly returns 404 when `req.Environment` names a non-existent env, but no test covers this branch. A negative test would prevent regressions on the validation logic.

PR #265, opened 2026-05-08
feat: add confirmation prompts for app/secret delete (#243)

### LINKED ISSUE #243: CLI UX: add confirmation prompts, deprecate --value flag
## Summary

CLI safety and security improvements.

## Details

**P3-15: app delete has no confirmation (app.go:142-162)**
\`app delete\` fires immediately with no confirmation prompt, no \`--yes\` flag, no interactive check. App deletion is destructive (garbage-collects Deployments, Services, Ingresses, PVCs, Secrets). Same issue with \`secret delete\`.

**P3-14: --value flag exposes secrets (secret.go:83)**
\`--value\` puts the secret in shell history, \`/proc/<pid>/cmdline\`, and \`ps\` output. The secure prompt via \`term.ReadPassword\` is the default but the flag's existence is a risk.

## Fix

1. Add interactive confirmation prompt for \`app delete\` and \`secret delete\`, with \`--yes\` flag for scripting
2. Deprecate \`--value\` flag with a warning message; recommend stdin pipe (\`echo \$SECRET | mortise secret set ...\`) or env var instead

MC-Meesh:
## Summary
- Add interactive confirmation prompts for destructive CLI operations (app delete, secret delete)
- Deprecate `--value` flag on secret set in favor of stdin/prompt input
- Add tests for confirmation flow

Closes #243

## Test plan
- [ ] `mortise app delete` prompts for confirmation
- [ ] `mortise secret delete` prompts for confirmation
- [ ] `--force` flag bypasses prompts
- [ ] CI pipeline green


MC-Meesh (commented):
**Code Quality Audit — PR #265**

**[NOTE]** cmd/cli/app.go:151 — `reader.ReadString('\n')` error is discarded (`line, _ := ...`). If stdin is a pipe/redirect that hits EOF without a newline, `line` will be empty and the command aborts (safe side), so this is not a bug. However, handling the error explicitly (e.g., treating read failure as abort) would make the intent clearer and avoid surprises if the logic changes later.

**[NOTE]** cmd/cli/secret.go:102 — Same pattern as above — discarded error from `reader.ReadString('\n')`. Same recommendation.

Overall: clean, consistent implementation. The `--yes` / `-y` flag follows standard CLI conventions, the deprecation warning on `--value` is a good security improvement, and tests cover the new flags.

MC-Meesh (commented):

**Findings:**

### WARNING
- **Swallowed error on stdin read.** In both `app.go` and `secret.go`, the `reader.ReadString('\n')` error is discarded with `line, _ := reader.ReadString('\n')`. If stdin is closed (piped context, CI, non-interactive shell), this silently returns an empty string, which fails the `EqualFold` check and aborts. This is arguably correct behavior (non-interactive = abort), but the user sees "Aborted." with no indication that it was because stdin was unavailable, not because they typed "n". Consider printing a more descriptive message when the error is non-nil (e.g., "Aborted: could not read from stdin (non-interactive shell? use --yes)").

### NOTE
- **`--value` deprecation warning fires even when value is empty.** The deprecation warning in `secret set` fires when `cmd.Flags().Changed("value")` is true, which is correct. But the subsequent `if secretValue == ""` block then prompts for interactive input, meaning a user who passes `--value ""` gets both a deprecation warning AND a terminal prompt. This is a minor UX wart — the deprecation message tells them to use stdin, but then the code falls through to `term.ReadPassword`. Not a bug, but worth a comment in the code explaining the interaction.

- **Tests only verify flag existence, not behavior.** `TestAppDelete_HasYesFlag` and `TestSecretDelete_HasYesFlag` check that the flag is registered with the correct shorthand, but don't test the actual confirmation flow (e.g., that stdin "n" aborts, that `--yes` skips the prompt). This is fine for a first pass but leaves the core behavior untested at the unit level.

MC-Meesh:
Follow-up audit for PR #265:

- The prior stdin-read concern is addressed in the current diff: app/secret delete and the existing project confirmations now return a descriptive error with a `--yes` hint when confirmation cannot be read.
- Local targeted verification passed: `go test ./cmd/cli`.
- Current GitHub checks: Unit + envtest, Helm lint/templates, Go vet/staticcheck, and UI build are green; Integration tests are still pending.
- Remaining actionable test gap: the added tests verify flag registration, but still do not cover delete confirmation behavior itself (`y` proceeds, non-`y` aborts, `--yes` skips the prompt). This is not a merge blocker by itself, but it leaves the core destructive-command safety behavior mostly untested.

PR #267, opened 2026-05-08
feat: improve UI handling of forbidden RBAC actions (#115)

### LINKED ISSUE #115: Improve UI handling of forbidden RBAC actions
## Problem

When a user lacks permission for an action (e.g., a member trying to delete an app or access admin settings), the UI lets them attempt it and then shows a generic "Forbidden" error banner. This is a poor experience — the user shouldn't be invited to click buttons that will fail.

## Current behavior

- All UI elements are rendered regardless of the user's role
- Forbidden actions return HTTP 403 from the API
- The UI displays a red error banner with the raw error message
- The user has no way to know ahead of time what they can and can't do

## Options to explore

**Option A: Hide forbidden actions entirely.** Query the user's permissions on page load and conditionally render buttons/sections. Admin-only areas (platform settings, user management, danger zones) simply don't appear for non-admins. Cleanest UX but requires the frontend to know the permission model.

**Option B: Show but disable with explanation.** Render forbidden actions as disabled/grayed-out elements with a tooltip like "Requires admin role." The user can see the capability exists but understands why they can't use it. More discoverable than hiding.

**Option C: Contextual error handling.** Keep the current render-everything approach but replace the generic "Forbidden" banner with a specific, styled message: "You don't have permission to delete apps. Contact a project admin." Less work but still a frustrating click-then-fail flow.

These aren't mutually exclusive — A or B for known permission boundaries (admin settings, delete actions) plus C as a fallback for edge cases.

## Where this matters most

- **Admin settings pages** (`/admin/settings`, `/settings/git-providers`) — members can navigate here and see forms they can't save
- **Danger zone** (delete app, delete project) — the delete button is visible to everyone
- **Project settings** (environments, members) — role-dependent actions mixed with read-only views
- **Deploy/redeploy buttons** — may be restricted by role

## Implementation considerations

- The API already returns the user's role in the auth response — the frontend has the info needed
- Permission model is simple in v1 (admin vs member), so the frontend logic is straightforward
- The authz package (`internal/authz/`) defines the full permission matrix — could expose a `/api/me/permissions` endpoint if needed
- Whatever approach is chosen should degrade gracefully: if the permission check fails, show the action and handle the 403 (option C)

MC-Meesh:
## Summary
- Improve UI error handling when users lack RBAC permissions
- Add graceful degradation for forbidden API responses in project settings
- Update store to track permission state for conditional UI rendering

Closes #115

## Test plan
- [ ] Users with restricted roles see appropriate UI state instead of errors
- [ ] Settings page degrades gracefully when user lacks write permissions
- [ ] CI pipeline green


MC-Meesh (commented):
**Code Quality Audit — PR #267**

**[WARNING]** ui/src/lib/store.svelte.ts:loadProjectRole — The `catch` block silently swallows all errors and returns `null`, which maps to "no role" and hides management UI. If a legitimate owner hits a transient network error during role loading, they will see a restricted view with no indication of why. Suggestion: distinguish between "not a member" (server returned 200 with no matching member) and "failed to load" (network/server error). For the error case, either surface a toast/banner ("Could not verify your role — some actions may be hidden") or default to showing the UI and relying on the server-side 403 as the enforcement layer.

**[NOTE]** ui/src/routes/projects/[project]/+page.svelte:128 — `loadProjectRole` runs in `Promise.all` with `listApps`, but if the role fetch is slower, there will be a brief flash where RBAC-gated sections are hidden then appear (or vice versa if the default state shows them). Consider showing a loading skeleton for gated sections until the role resolves, or ensuring `loadProjectRole` completes before rendering gated content.

**[NOTE]** ui/src/lib/api.ts:54 — The 403 handler is a good addition. One edge case: if the server returns 403 with a non-JSON body, `res.json().catch()` correctly falls back to `{ error: 'forbidden' }`. This is handled correctly.

The overall approach — client-side gating for UX + server-side 403 for enforcement — is the right architecture. The main concern is the silent failure mode in role loading.

MC-Meesh (commented):

**Findings:**

### WARNING
- **Role loading is fire-and-forget in `Promise.all` — no error boundary for the role half.** In both `+page.svelte` (project view) and `settings/+page.svelte`, the `loadProjectRole` call is inside `Promise.all` alongside `api.listApps` / `api.getProject`. The fix commit correctly made `loadProjectRole` catch internally and surface a warning banner instead of throwing. However, if a future change accidentally removes that internal catch, `Promise.all` would reject and the entire page load would fail (showing the catch block's generic error). Consider adding a comment near the `Promise.all` calls noting that `loadProjectRole` is intentionally non-throwing, or use `Promise.allSettled` for defense in depth.

- **`canManageMembers` and `canDeleteInProject` are identical.** Both methods return `true` only for admin or `owner` role. If the intent is that developers should eventually be able to delete apps (but not projects or members), having two identically-implemented methods is good forward-looking design. But if that distinction is never expected, the duplication is misleading. Worth a brief code comment documenting the intended future divergence (or collapsing them if none is planned).

### NOTE
- **403 handler consumes response body, preventing downstream handling.** The new 403 block in `api.ts` calls `res.json()` and throws before the existing `!res.ok` block. This is correct — but note that `res.json()` consumes the body stream. If a future refactor reorders these blocks or adds a fallthrough, the second `res.json()` call would fail. The current code is fine, just fragile to reordering.

- **`loadProjectRole` caches but never invalidates.** Once a role is loaded for a project, it stays in `projectRoles` for the lifetime of the store (page session). If an admin changes a user's role while they have the page open, the UI won't reflect it until a full page reload. This is acceptable for a v1 but worth noting for future SSE/websocket role-change events.

- **Platform settings page git provider delete gate.** The `{#if store.isAdmin}` guard on the delete button in `settings/+page.svelte` is correct — platform-level settings should be admin-only. But the rest of the git provider management UI (add, edit) is not similarly gated. If a non-admin can somehow reach `/settings`, they could add providers but not delete them, which is an inconsistent permission surface. This may already be gated by route-level auth, but worth verifying.

PR #268, opened 2026-05-08
fix: make PATCH fields clearable and batch env build status updates (#211)

### LINKED ISSUE #211: Non-clearable PATCH fields and status update conflicts
## Context

During the review and fix-up of PR #209 (per-env builds, domain UX, health fixes), several follow-up issues were identified that are out of scope for the current PR but should be addressed.

## 1. Non-clearable PATCH fields in platform config

**Severity:** Low — only matters if a user needs to revert a setting to empty.

The `buildPlatformSpec` function in `internal/api/platform.go` uses `!= ""` guards on string fields, which means once a value is set, sending an empty string in the PATCH body can't clear it. PR #209 fixed this for `ExternalDomain` by converting it to `*string`, but the same pattern exists for:

- `Domain`
- `DomainTemplate`
- `TLS.CertManagerClusterIssuer`
- `Registry.URL`, `Registry.Namespace`
- `Build.BuildkitAddr`, `Build.DefaultPlatform`
- `Observability.LogsAdapterEndpoint`, `MetricsAdapterEndpoint`, `TrafficAdapterEndpoint`

**Partially fixed:** `ExternalDomain` was converted to `*string` with `!= nil` / dereference in PR #209. The remaining fields listed above still use the old `!= ""` pattern.

**Fix:** Convert each remaining field that should be clearable to `*string` with the same `!= nil` / dereference pattern used for `ExternalDomain`. Not all fields necessarily need to be clearable (e.g., `Domain` probably shouldn't be empty), so this should be evaluated per-field.

## 2. Status update conflicts in per-env build loop

**Severity:** Low — the build result still flows correctly; only the status write is transiently lost.

`applyEnvBuildSuccess` in `internal/controller/app_controller.go` calls `r.Status().Update(ctx, app)` inside the env loop. When processing multiple environments in a single Reconcile call, the second status update may conflict with the first (stale `resourceVersion`). The error is logged but discarded, and the image is still returned to the caller, so deployments proceed correctly. The status eventually converges on the next reconcile cycle.

**Fix options:**
- Batch all env status updates into a single `Status().Update()` after the loop
- Switch to `Status().Patch()` which doesn't require the full object version
- Accept the current behavior since it self-heals (document the trade-off)

## 3. `allEnvBuildsCurrentForRevision` checks app spec, not resolved envs

**Severity:** Very low — causes one extra unnecessary reconcile cycle in a rare edge case.

The `allEnvBuildsCurrentForRevision` gate checks `app.Spec.Environments` to decide if all builds are current and `prepareGitSource` can be skipped. However, `resolveEnvs` may produce additional environments from the Project spec that aren't in the App spec. If a new Project environment is added, the gate won't detect it as needing a build. The build will fail on the first attempt (no cached git credentials), then succeed on the next reconcile when `prepareGitSource` runs.

**Fix:** Move project fetch + `resolveEnvs` before the git-source switch statement, then pass the resolved env list to the gate. This is a minor refactor of the reconcile flow.

---

All three issues are non-blocking and self-healing. They can be addressed in a follow-up PR.

MC-Meesh:
## Summary
- Fix non-clearable PATCH fields in platform API — allow null/empty values to clear fields
- Batch environment build status updates to reduce reconcile conflicts

Closes #211

## Test plan
- [ ] PATCH with null values correctly clears fields
- [ ] Build status updates are batched without conflicts
- [ ] CI pipeline green


MC-Meesh (commented):
**Code Quality Audit — PR #268**

**[WARNING]** `internal/api/platform.go` — Observability token fields (`LogsAdapterToken`, `MetricsAdapterToken`, `TrafficAdapterToken`) are intentionally left as `string` (not `*string`), which means they cannot be cleared via PATCH with an empty string — the `omitempty` tag will drop them. The endpoint fields were correctly converted to `*string`. If clearing adapter tokens is a valid use case (e.g., disconnecting an observability backend), these should also be `*string`. If this is intentional (tokens are write-only / never clearable), add a comment explaining why they differ from their sibling endpoint fields.

**[WARNING]** `internal/controller/app_controller.go` — The batched status flush only runs when `app.Spec.Source.Type == SourceTypeGit && r.BuildClient != nil`. If `applyEnvBuildSuccess` is ever called for a non-git source type (or when BuildClient is nil due to config), the in-memory status mutations will be silently dropped since the per-env `Status().Update()` was removed. Verify that `applyEnvBuildSuccess` is exclusively reachable under those same conditions.

**[NOTE]** `internal/controller/app_controller.go` — The batched flush logs the error but does not return it or set a requeue. If the single `Status().Update()` fails (e.g., conflict), all env build statuses accumulated in that reconcile loop are lost until the next reconcile. Consider returning a `ctrl.Result{Requeue: true}` on failure so the status is retried promptly.

MC-Meesh (commented):

**Findings:**

### WARNING
- **Adapter token deletion leaves stale secret behind.** When all three adapter tokens are cleared (sent as empty strings), `upsertAdapterTokens` deletes the individual keys from the Secret via `delete(secret.Data, ...)` but never checks whether the Secret is now empty. An empty Secret persists as an orphan in the cluster. Consider deleting the Secret entirely when `len(secret.Data) == 0` after the key removals, or add a comment documenting why the empty Secret is intentional.

- **Build status flush error is swallowed silently.** The new batched `Status().Update()` after the env loop logs the error but does not set a failed condition on the App or trigger a requeue. If this write fails (e.g. due to a stale resourceVersion), the in-memory status diverges from etcd and no retry occurs until the next external reconcile trigger. Consider returning `ctrl.Result{Requeue: true}` on error here so the controller self-heals.

### NOTE
- **Variable naming: `certGVR` is actually a `GroupVersionKind`.** Not in this PR — just flagging that the variable introduced in a sibling PR (#270) uses `GVR` but holds a `schema.GroupVersionKind`. If these PRs merge together, a rename to `certGVK` would avoid confusion.

- **`buildStatusDirty` won't catch the path where `reconcileEnvBuild` returns early with a cached image.** When the env already has the correct `lastBuiltRevision`, `reconcileEnvBuild` returns `(es.LastBuiltImage, false, false)` — `statusDirty=false`. This is correct because no status mutation occurred, but worth confirming that the only caller of `applyEnvBuildSuccess` is the `buildPhaseSucceeded` path (which does return `dirty=true`). Verified — this is fine.

PR #269, opened 2026-05-08
feat: JWT token refresh flow (#251)

### LINKED ISSUE #251: No JWT token refresh flow — expired tokens require full re-login
## Summary

JWT tokens have a 24-hour TTL (`internal/auth/jwt.go:21`) with no refresh mechanism. When a token expires, the only path back is `mortise login`. The CLI does not detect 401 responses and prompt for re-authentication — it just prints a raw error.

## Current behavior

- Token generated at login with 24h expiry
- No `/api/auth/refresh` endpoint exists
- CLI `doJSON()` returns raw `"API error 401: ..."` without special handling
- No refresh token is issued alongside the access token
- CI/CD pipelines using API tokens silently break after 24 hours

## Expected behavior

At minimum:
1. CLI should detect 401 and prompt for re-login (or auto-refresh if a refresh token exists)
2. A refresh endpoint (`POST /api/auth/refresh`) should allow extending sessions without re-entering credentials
3. For CI/CD: support long-lived API tokens (deploy tokens) separate from user session tokens

## Files involved

- `internal/auth/jwt.go` — token generation, 24h TTL
- `internal/api/auth.go` — login endpoint (no refresh endpoint)
- `internal/api/middleware.go` — JWT validation middleware
- `cmd/cli/client.go:105-107` — raw error return on 401
- `cmd/cli/login.go` — login flow

MC-Meesh:
## Summary
- Add JWT token refresh endpoint for seamless session extension
- Implement client-side automatic refresh before token expiry
- Add native auth refresh handler

Closes #251

## Test plan
- [ ] Expired tokens trigger refresh flow instead of re-login
- [ ] Refresh endpoint returns valid new JWT
- [ ] CI pipeline green


MC-Meesh (commented):
**Code Quality Audit — PR #269**

**[WARNING]** `ui/src/lib/api.ts` — Race condition on concurrent 401s. If multiple `request()` calls are in-flight when the token expires, each one independently calls `tryRefreshToken()`. The first succeeds and writes the new token to localStorage, but subsequent calls send the *old* token to `/auth/refresh` and may fail (or worse, each gets a different new token, causing the earlier refreshed requests to retry with a stale token). Consider a shared refresh promise (e.g., a module-level `let refreshPromise: Promise<boolean> | null`) so concurrent callers coalesce into a single refresh.

**[WARNING]** `cmd/cli/client.go:doJSON` — The original response body is read/closed before the refresh attempt, but if `tryRefreshToken()` returns false, execution falls through to the `resp.StatusCode >= 400` block which calls `io.ReadAll(resp.Body)` on the *already-closed* original response body. This will return an empty body, so the error message becomes `"API error 401: "` instead of showing the server's error message. The body should be read before closing it for the refresh path, or the original bytes should be preserved.

**[NOTE]** `internal/api/server.go` — The `/api/auth/refresh` endpoint is listed as unauthenticated (no middleware), which is correct since it accepts expired tokens. However, this means it is also exposed to unauthenticated callers with no token at all. The handler does check for a bearer token, but consider rate limiting this endpoint to prevent brute-force token guessing (even though JWTs are hard to guess, the endpoint still does crypto work per request).

**[NOTE]** `internal/auth/jwt.go` — The 7-day refresh window is generous. A stolen token remains usable for a full week after expiry. This is documented in the swagger spec, so it appears intentional — just flagging it as a security trade-off to be aware of.

MC-Meesh (commented):

**Findings:**

### CRITICAL
- **Refresh endpoint does not enforce OIDC-path invalidation.** The `Refresh` handler type-asserts `s.auth` to `*auth.NativeAuthProvider` and calls `CheckPasswordGen` only when the assertion succeeds (`ok == true`). When the platform uses OIDC auth, the assertion fails silently and `ok == false`, so the password-generation check is **skipped entirely**. This means an OIDC user whose account is revoked/disabled on the IdP can keep refreshing their JWT indefinitely (up to the 7-day window) with no server-side revocation check. At minimum, when `!ok` and the auth provider is non-native, the refresh should be denied or an equivalent revocation check must exist for the OIDC path.

### WARNING
- **`tryRefreshToken` in CLI sends the refresh request using `c.do()`, which attaches the current (expired) token via the Authorization header.** If `c.do()` itself triggers a 401 → refresh loop in a future refactor (e.g. if `doRaw` gains retry logic), this could cause infinite recursion. Consider adding a `skipRetry` flag or using a raw `http.Client.Do()` for the refresh call to make the non-recursive intent explicit.

- **UI refresh deduplication has a race window.** `refreshPromise` is a module-level `let` shared across concurrent `request()` calls. If the first call's `tryRefreshToken()` resolves to `false`, the `.finally()` nulls out `refreshPromise`, and a second in-flight call that was `await`-ing the same promise correctly gets `false`. However, a **third** call arriving between `finally()` and the redirect will start a new refresh attempt against an already-failed token, resulting in a wasted round-trip. This is low severity but could be tightened by caching the `false` result for a short window.

- **7-day refresh window is generous.** A stolen JWT that has expired can be refreshed for a full week. The `CheckPasswordGen` guard only protects native-auth users who change their password. Consider documenting this trade-off or making the refresh window configurable (e.g. via PlatformConfig).

### NOTE
- **`resp.Body` lifecycle in `doJSON` is correct but fragile.** The `defer resp.Body.Close()` is positioned after the 401 early-return block, so it is never registered on the 401 path (where the body is manually drained and closed). This works today but is easy to break if someone reorders the code. A brief comment would help.

JamesSprow (dismissed):
Re-audited the current head. No further correctness or regression findings beyond the client retry fix already pushed.

JamesSprow (dismissed):
Re-audited the updated head after merging main. No remaining correctness findings; local validation is green and the merged refresh/auth paths are consistent.

JamesSprow (approved):
Re-audited the post-fix head. The duplicated test helper merge regression is resolved; local vet and targeted local tests pass.

PR #270, opened 2026-05-08
feat: cert-manager integration — ClusterIssuer, cert status, admin TLS (#80)

### LINKED ISSUE #80: cert-manager integration: ClusterIssuer default, cert status surfacing, admin TLS
### Gap 1 — No ClusterIssuer default
PlatformConfig has a `tls.certManagerClusterIssuer` field but no default is created. Users must manually create a ClusterIssuer and configure it. quick-mortise (#72) should handle this automatically.

### Gap 6 — Custom domain cert status
When a user adds a custom domain, cert-manager's HTTP-01 or DNS-01 challenge may fail (DNS not pointed, port 80 blocked). Mortise has no visibility — the user just sees HTTPS not working. Need to:
- Watch cert-manager Certificate resources (conditional — cert-manager may not be installed)
- Surface challenge status in App status / UI ("Certificate pending — point DNS to X")

### Gap 7 — Admin Ingress TLS
Mortise's own Helm chart Ingress has no cert-manager annotation. Admin UI runs HTTP. Fix: add the annotation to the chart template, gated on `tls.certManagerClusterIssuer` being set.

## Testing
Requires a real domain with DNS pointing at the cluster. Can use Cloudflare DNS-01 challenges (no port 80 needed) or HTTP-01 (requires port 80 accessible). Local k3d testing not possible without domain setup.

## Related
- #77 (routing/ingress RFC — gaps 1, 6, 7)
- #72 (quick-mortise — would auto-create ClusterIssuer)

MC-Meesh:
## Summary
- Add default ClusterIssuer support for automatic TLS certificate provisioning
- Surface certificate status in app/environment status
- Enable admin TLS configuration via CRDs and Helm values

Closes #80

## Test plan
- [ ] ClusterIssuer is created when cert-manager is enabled
- [ ] Certificate status is reflected in app status
- [ ] CI pipeline green


MC-Meesh (commented):
**Code Quality Audit — PR #270**

**[WARNING]** `internal/controller/app_controller.go:updateStatus` — When `env.TLS.SecretName` is set (user-provided TLS secret), the code sets `tlsSecret = ""` and skips certificate status checking entirely. This means apps using a custom TLS secret with cert-manager will never show certificate status in the UI. The intent seems to be "skip cert-manager status for manually-managed secrets," but if the custom secret is still backed by a cert-manager Certificate, the status is silently lost. Consider using the custom SecretName as the Certificate lookup key instead of blanking it out.

**[WARNING]** `internal/controller/app_controller.go:updateStatus` — The TLS secret name is derived as `fmt.Sprintf("%s-tls", app.Name)`, but this must match whatever name cert-manager actually uses for the Certificate resource (typically derived from the Ingress TLS block's `secretName`). If the Ingress reconciler uses a different naming convention, this lookup will silently return ("", "") and cert status will never appear. Ensure this name is consistent with `reconcileIngress`.

**[NOTE]** `internal/controller/app_controller.go` — The variable `certGVR` is actually a `schema.GroupVersionKind` (GVK), not a GVR. This is cosmetic but will confuse future readers. Consider renaming to `certGVK`.

**[NOTE]** `charts/mortise/templates/clusterissuer.yaml` — The `email` and `server` values are injected without Helm `quote`. While these come from values.yaml (not user input in most cases), quoting them (`{{ .Values.tls.acme.email | quote }}`) would be defensive against YAML injection if values are templated from external sources.

MC-Meesh (commented):

**Findings:**

### WARNING
- **Variable named `certGVR` holds a `schema.GroupVersionKind`, not a GVR.** `var certGVR = schema.GroupVersionKind{...}` is misleading. GVR (GroupVersionResource) and GVK (GroupVersionKind) are distinct concepts in the k8s type system. Rename to `certGVK` to match its actual type and avoid confusion for future maintainers.

- **`checkCertificateStatus` returns "Failed" for any non-True Ready condition.** A cert-manager Certificate with `Ready=False` and reason `Issuing` (in-progress renewal) will be reported as "Failed" to the user when it is actually still being processed. The mapping should distinguish between transient reasons (e.g. `Issuing`, `Pending`) and terminal failures (e.g. `InvalidCertificate`, `OrderFailed`). At minimum, consider treating `Ready=False` as "Pending" unless the reason explicitly indicates a terminal failure.

- **Admin Ingress TLS template puts annotation on the Ingress `metadata` (outside `spec`).** The template in `deployment.yaml` adds `cert-manager.io/cluster-issuer` as an annotation on the Ingress metadata block, which is correct for cert-manager's annotation-based integration. However, note that this is technically a cert-manager-specific annotation on a standard Ingress resource. The CLAUDE.md architecture rules say "k8s Ingress API: not Traefik-specific annotations" but cert-manager annotations are industry-standard and expected. Flagging for awareness rather than as a violation.

### NOTE
- **No watch on cert-manager Certificate resources.** `checkCertificateStatus` reads Certificate status during each reconcile cycle, but the controller does not watch Certificate resources for changes. This means certificate status changes (e.g. issuance completing) will only be reflected on the next App reconcile trigger, not in real-time. Consider adding a watch with an `EnqueueRequestForOwner` mapper if cert-manager Certificates are owned by the Ingress (which they typically are via the annotation integration). Low priority since reconcile loops run frequently.

- **`clusterissuer.yaml` hardcodes `http01.ingress: {}`.** This uses cert-manager's default ingress solver, which works for most setups. However, if the cluster uses a non-default ingress class, challenges may fail. The umbrella chart could expose `tls.acme.solverIngressClass` as an optional value for non-standard setups. Low priority — the current default is reasonable.

MC-Meesh:
Audit follow-up for PR #270:

No blocking behavioral regression found in the cert-manager/ClusterIssuer, controller status, RBAC, OpenAPI/CRD, or UI banner changes I reviewed.

Verification:
- `make test-charts` passed locally.
- GitHub checks currently show Unit + envtest, Helm lint/template, Go vet/staticcheck, and UI typecheck/build passing.
- Integration tests are still pending, and the PR remains blocked on required review.
- A direct local `go test ./internal/controller ...` did not run because envtest binaries are unavailable in this worktree (`../../bin/k8s`, `/usr/local/kubebuilder/bin/etcd`), but the PR's Unit + envtest check passed.

Actionable test gap before merge: the diff adds nuanced certificate status behavior (`Ready`/`Pending`/`Failed` mapping and custom TLS secret validation) without changed tests/specs. I recommend adding focused controller tests for cert-manager condition mapping and BYO TLS secret Ready/Failed status, or explicitly accepting that coverage gap.

PR #272, opened 2026-05-08
fix: SSE token exchange, JWT refresh, and client-side role trust (#232)

### LINKED ISSUE #232: JWT token exposed in SSE URL + no refresh handling
## Summary

The UI passes the JWT bearer token as a URL query parameter for SSE endpoints, and has no token refresh or expiry handling.

## Details

**P1-3: JWT in SSE URL (api.ts:239)**

```typescript
if (token) params.set('token', token);
```

The JWT is exposed in browser address bars, HTTP Referer headers, server access logs, proxy logs, and browser history. SSE/EventSource doesn't support custom headers, which motivates this pattern, but the standard mitigation is a short-lived scoped token exchange.

**P1-14: No token refresh (api.ts:36-68)**

`request()` handles 401 by redirecting to login, but there's no proactive token refresh, no expiry check, and no refresh token flow. The token in localStorage lives indefinitely until a 401 happens. Active sessions drop abruptly.

**P4-8: JWT decoded without signature verification (store.svelte.ts:71-76)**

The JWT payload is decoded client-side with `atob` and used to seed user identity including `role`. Backend auth still blocks, but UI-level access control is bypassed by any crafted token in localStorage.

## Discussion needed

For the SSE token issue, the standard fix is a short-lived token exchange endpoint:
1. Client calls `POST /api/auth/sse-token` with JWT in Authorization header
2. Server returns a single-use, short-TTL token scoped to SSE
3. Client passes this token in the SSE URL

Is this the right approach, or is there a simpler alternative given the deployment model (typically behind an auth proxy)?

MC-Meesh:
## Summary
- Replace exposed JWT in SSE URL with short-lived token exchange
- Add JWT refresh handling on the client side
- Remove client-side role trust — verify roles server-side

Closes #232

## Test plan
- [ ] SSE connections use exchanged tokens, not raw JWTs
- [ ] Token refresh works seamlessly during long sessions
- [ ] CI pipeline green

### DIFF
--- a/cmd/observer/collectors_test.go
+++ b/cmd/observer/collectors_test.go
@@ -3,8 +3,11 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"io"
 	"log/slog"
 	"math"
+	"net/http"
 	"path/filepath"
 	"testing"
 	"time"
@@ -13,7 +16,12 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/kubernetes/scheme"
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	rest "k8s.io/client-go/rest"
+	fakerest "k8s.io/client-go/rest/fake"
 	clienttesting "k8s.io/client-go/testing"
 	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	metricsfake "k8s.io/metrics/pkg/client/clientset/versioned/fake"
@@ -26,6 +34,53 @@ func discardLogger() *slog.Logger {
 type nopWriter struct{}
 
 func (nopWriter) Write(p []byte) (int, error) { return len(p), nil }
+
+type blockingLogClientset struct {
+	kubernetes.Interface
+	t *testing.T
+}
+
+func (c *blockingLogClientset) CoreV1() coreclientv1.CoreV1Interface {
+	return &blockingLogCoreV1{CoreV1Interface: c.Interface.CoreV1(), t: c.t}
+}
+
+type blockingLogCoreV1 struct {
+	coreclientv1.CoreV1Interface
+	t *testing.T
+}
+
+func (c *blockingLogCoreV1) Pods(namespace string) coreclientv1.PodInterface {
+	return &blockingLogPods{PodInterface: c.CoreV1Interface.Pods(namespace), namespace: namespace, t: c.t}
+}
+
+type blockingLogPods struct {
+	coreclientv1.PodInterface
+	namespace string
+	t         *testing.T
+}
+
+func (p *blockingLogPods) GetLogs(name string, opts *corev1.PodLogOptions) *rest.Request {
+	reader, writer := io.Pipe()
+	p.t.Cleanup(func() {
+		_ = reader.Close()
+		_ = writer.Close()
+	})
+
+	fakeClient := &fakerest.RESTClient{
+		Client: fakerest.CreateHTTPClient(func(_ *http.Request) (*http.Response, error) {
+			return &http.Response{StatusCode: http.StatusOK, Body: reader}, nil
+		}),
+		NegotiatedSerializer: scheme.Codecs.WithoutConversion(),
+		GroupVersion:         corev1.SchemeGroupVersion,
+		VersionedAPIPath:     fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/log", p.namespace, name),
+	}
+	return fakeClient.Request()
+}
+
+func withBlockingPodLogs(t *testing.T, cs kubernetes.Interface) kubernetes.Interface {
+	t.Helper()
+	return &blockingLogClientset{Interface: cs, t: t}
+}
 
 // --- parseLogTimestamp ---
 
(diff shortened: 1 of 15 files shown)


MC-Meesh (commented):
**[WARNING]** ui/src/lib/api.ts:tokenExpiresAt — `atob()` does not handle base64url encoding. JWTs use base64url (with `-` and `_` characters), but `atob()` expects standard base64 (with `+` and `/`). If the JWT payload contains these characters, `atob()` will throw and `tokenExpiresAt()` returns null, silently disabling token refresh. Fix: replace `atob(token.split('.')[1])` with a base64url-safe decode, e.g. `atob(token.split('.')[1].replace(/-/g, '+').replace(/_/g, '/'))`.

**[WARNING]** internal/api/sse_tokens.go:RefreshToken — The refresh endpoint issues a new JWT for any valid JWT, with no secondary verification (e.g., checking a token family/generation, or requiring a separate refresh token). A stolen JWT can be indefinitely refreshed to maintain access. Consider: (a) adding a refresh token rotation scheme, or (b) at minimum, checking that the user still exists and is not disabled before issuing a new token.

**[NOTE]** internal/api/sse_tokens.go:sseTokenStore — The in-memory token store means SSE tokens issued by one server replica cannot be redeemed by another. This is fine for single-replica deployments, but will break if the operator is scaled horizontally. Worth documenting this constraint or planning for a shared store (Redis, etc.) before scaling.

**[NOTE]** internal/api/sse_tokens.go:IssueSSEToken + RefreshToken — Neither endpoint has rate limiting. An attacker with a valid JWT could flood the SSE token store or generate refresh tokens at high volume. Not critical if the JWT auth middleware provides sufficient protection, but worth considering per-user rate limits on these endpoints.

MC-Meesh (commented):

**Findings:**

### WARNING

1. **`time.Now()` in SSE token store (sse_tokens.go lines 51, 69, 82):** The `sseTokenStore` calls `time.Now()` directly in `Issue()`, `Redeem()`, and `cleanup()`. The CLAUDE.md architecture rules state: "Never use `time.Now()` in controllers — use injected clock." While `sseTokenStore` lives in the API layer rather than a controller, it still makes testing TTL/expiration behavior difficult without a fake clock. Since the cleanup goroutine runs on a real ticker, tests cannot exercise expiration paths deterministically. Consider accepting a `clock.Clock` parameter (same pattern the controllers use) for testability.

2. **`RefreshToken` re-issues JWT without re-validating the user:** The `RefreshToken` handler (sse_tokens.go ~line 133) takes the principal from the JWT middleware and immediately mints a new JWT. If a user was disabled or had their role changed after the original token was issued, the refreshed token will carry stale claims until the original token expires. A refresh endpoint is a natural place to re-read the user from the auth provider to pick up changes. Not a security hole (the old token is still valid for its remaining lifetime anyway), but it weakens the security posture that the refresh feature is meant to improve.

3. **SSE token single-use not tested through the middleware:** `TestSSETokenRedeemAndSingleUse` only verifies that two issued tokens are different and have the right prefix. It does not actually redeem a token through `sseAuthMiddleware` and then verify that a second use of the same token returns 401. The test name implies single-use validation but does not exercise it end-to-end.

### NOTE

1. **`logsURL` is now async — callers must await:** The change from `logsURL(): string` to `logsURL(): Promise<string>` is a breaking API contract change. All callers in this PR (`LogViewer.svelte`, `DeployLogsTab.svelte`, `projectEvents.ts`, `+page.svelte`) have been updated, which is correct. However, any external or yet-unmerged code that calls `api.logsURL` synchronously will silently get a Promise object as the URL string, which will fail at runtime without a compile-time error if the caller doesn't type-check the return. Flagging for awareness rather than as a bug.

2. **Test fixture annotations cross PR boundary:** Every test fixture got a `mortise.dev/disable-default-security-context: "true"` annotation, which references functionality from PR #275 (Helm security contexts). If #272 merges before #275, these annotations will be inert (no code reads them yet). If #275 merges first, the test fixtures without the annotation may break. Coordinate merge order or ensure the annotation-reading code degrades gracefully.

3. **`jwtAuthMiddleware` early return for already-authenticated requests:** The new guard at the top of `jwtAuthMiddleware` (`if PrincipalFromContext(r.Context()) != nil { next.ServeHTTP... }`) is correct for the SSE flow (where `sseAuthMiddleware` already injected a principal). However, it means that if any middleware earlier in the chain erroneously sets a principal in the context, `jwtAuthMiddleware` will skip JWT validation entirely. This is safe given the current middleware ordering, but worth a comment explaining the contract.

MC-Meesh:
Audit follow-up for PR #272:

Findings that still look actionable on the current head (`ea6462e`):

1. `internal/api/sse_tokens.go:RefreshToken` still does not fully revalidate current user state. It calls `s.auth.Principal(...)`, but `NativeAuthProvider.Principal` validates the JWT and only checks `password_gen` if the user secret exists. If the user secret was deleted/revoked, the lookup miss is ignored and the stale JWT principal is returned. If the user's platform role changed, `Principal` also returns the role from JWT claims rather than reloading `secret.Data["role"]`. That means `/api/auth/refresh` can mint a fresh 24h JWT for a revoked user, or preserve a stale role across refreshes. The earlier refresh warning is only partially addressed.

2. `ui/src/lib/api.ts:fetchSSEToken` bypasses `maybeRefreshToken()`. Normal API calls refresh near expiry, but SSE setup paths (`logsURL`, `connectProjectEvents`) can be the first network call after a long idle session and will request `/auth/sse-token` using the old JWT directly. If the JWT is expired, the SSE token request fails and the EventSource opens without `token`, producing a 401/reconnect loop instead of refreshing first.

3. `git diff --check origin/main...HEAD` still fails:
   - `internal/api/sse_tokens.go:168: new blank line at EOF.`
   - `internal/api/sse_tokens_test.go:133: new blank line at EOF.`

4. `TestSSETokenRedeemAndSingleUse` still does not redeem a token through the SSE middleware or assert that a second use returns 401. It currently verifies uniqueness/prefix only, so the single-use behavior remains untested end to end.

Status checked: branch contains current `origin/main`; mergeable but blocked by required review and Integration tests still in progress while the other listed checks are green.

JamesSprow (approved):
Re-audited the current head. No further correctness or regression findings.

PR #273, opened 2026-05-08
fix: CI/CD supply chain hardening (#227)

### LINKED ISSUE #227: CI/CD supply chain: no image signing, broken workflow_dispatch, unverified binaries
## Summary

Three supply chain security issues in the GitHub Actions workflows.

## Details

**P0-19: No image signing or provenance (release.yml:54-74)**

`docker/build-push-action@v5` pushes images with no cosign signing, no SBOM, no SLSA provenance attestation. Consumers cannot verify image authenticity.

**P0-20: workflow_dispatch produces broken releases (release.yml:7,79,99,162)**

`workflow_dispatch` is enabled but chart/release jobs have `if: startsWith(github.ref, 'refs/tags/v')` guards. A manual dispatch from `main` pushes an image tagged with the branch name but skips chart packaging and GitHub Release creation, leaving orphaned images in GHCR.

**P0-21: k3d binary downloaded with no checksum (ci.yml:73-74)**

```yaml
curl -fsSL -o /usr/local/bin/k3d "https://github.com/k3d-io/k3d/releases/download/v5.8.3/k3d-linux-amd64"
chmod +x /usr/local/bin/k3d
```

No `sha256sum --check`, no GPG verification. A compromised release or MITM injects a malicious binary with full CI access.

## Fix

1. Add cosign signing and SLSA provenance to the build-push step
2. Either remove `workflow_dispatch` or add the tag guard to image build steps
3. Add `sha256sum --check` after k3d download with a pinned hash

MC-Meesh:
## Summary
- Fix broken workflow_dispatch in release workflow
- Add image signing step to CI pipeline

Closes #227

## Test plan
- [ ] CI workflow runs successfully
- [ ] Release workflow dispatch works
- [ ] CI pipeline green


MC-Meesh (commented):
**[WARNING]** .github/workflows/release.yml — The `workflow_dispatch` trigger was removed. This means releases can no longer be triggered manually, only by the remaining triggers (presumably tag push). If there's ever a need to re-run a failed release or do an out-of-band release, there's no manual escape hatch. Suggest keeping `workflow_dispatch` alongside the existing triggers.

**[NOTE]** .github/workflows/release.yml:cosign sign — The cosign signing step uses `--yes` to skip confirmation, which is correct for CI. The OIDC-based keyless signing via `id-token: write` permission is the right approach. The overall supply chain hardening (checksum verification of k3d, cosign image signing, SBOM generation, provenance attestation) looks solid.

MC-Meesh (commented):

**Findings:**

### CRITICAL

1. **Cosign actions not pinned to digest:** The `sigstore/cosign-installer@v3` action is referenced by mutable tag, not by SHA digest. This is exactly the class of supply chain vulnerability this PR aims to address. A compromised tag could inject a malicious cosign binary that signs images with an attacker-controlled key (or exfiltrates the OIDC token). Pin to a specific commit SHA, e.g. `sigstore/cosign-installer@dc72c7d5c4d10cd6bcb8cf6e3fd1d5ebbe264f12` (or whatever the current v3 HEAD is). The same principle applies to all other uses of `@v5`, `@v4`, etc. in the file — this PR should either pin them all or document why it does not.

### WARNING

1. **`id-token: write` and `attestations: write` appear at the workflow level, not the job level:** These permissions are granted to every job in the release workflow, including jobs that do not need them (e.g., the CLI build job, the chart publish job). The principle of least privilege says these should be scoped to the `image` job only. If a future job is added to the workflow, it will inherit these elevated permissions silently.

2. **No verification step after signing:** The workflow signs both images but never verifies the signatures. Adding a `cosign verify` step after signing (using the identity-based keyless verification) would catch signing failures before the release is published, and serves as documentation of the expected verification command for users.

3. **k3d checksum validation uses a fixed hash but no version pin:** The CI workflow now validates the k3d binary checksum, which is good. However, the download URL for k3d is not visible in this diff — if the URL uses a mutable tag (e.g., `latest`), then the checksum will fail when k3d releases a new version, breaking CI. Conversely, if the URL pins a version, the checksum is redundant with HTTPS integrity (though still a defense-in-depth measure). Verify the download URL pins the exact version that matches this hash.

### NOTE

1. **`provenance: mode=max` and `sbom: true` are good additions.** These add build provenance attestation and SBOM generation to the image build, which complements the cosign signing well.

MC-Meesh (commented):
## Audit follow-up

Findings from the current branch (`polecat/mutant/mo-eiy@mowmmpb8`):

### WARNING

1. The supply-chain hardening is still partial because most GitHub Actions remain referenced by mutable major tags. `sigstore/cosign-installer` is now pinned to a commit SHA, but this workflow still uses mutable refs such as `actions/checkout@v4`, `docker/setup-qemu-action@v3`, `docker/setup-buildx-action@v3`, `docker/login-action@v3`, `docker/metadata-action@v5`, `docker/build-push-action@v5`, `azure/setup-helm@v4`, `actions/setup-go@v5`, `actions/upload-artifact@v4`, `actions/download-artifact@v4`, and `softprops/action-gh-release@v2`. CI also has mutable action refs. If the goal of #227 is CI/CD supply-chain hardening, this should either pin the remaining third-party actions to immutable SHAs or explicitly document why only the newly-added cosign installer is pinned.

### Status

- `workflow_dispatch` is present in the final diff, so the manual release escape hatch is restored.
- `id-token` and `attestations` are scoped to the `image` job only; that addresses the earlier least-privilege concern for OIDC/attestation permissions.
- The k3d URL is pinned to `v5.8.3`, and the added checksum matches the pinned-version pattern.
- Current checks: Unit/envtest, Helm, Go vet/staticcheck, and UI build are passing; Integration tests are still pending, so I cannot call CI fully green yet.

PR #274, opened 2026-05-08
feat: per-env stale indicators and redeploy on project page (#213)

### LINKED ISSUE #213: Project page per-app redeploy button always redeploys all stale envs
## Current behavior

The project page has a stale-apps banner at the bottom of the canvas. Each stale app gets a button, plus there's a "Redeploy all" button. Both now call `redeploy-stale` (POST `/api/.../redeploy-stale`), which restarts **all** stale environments for the app.

This is correct for multi-env apps where the user wants to apply all pending changes at once — and it fixes the previous bug where `redeployOne` passed the project page's `selectedEnv` regardless of which envs were actually stale.

## Potential improvement

In some workflows a user might want to redeploy only one specific stale env from the project page without opening the drawer. Two options:

1. **Show per-env stale indicators on the app card** — e.g. a small badge per stale env, each clickable to redeploy just that env (calls existing `api.redeploy(project, app, env)`).
2. **Keep current behavior** — the per-app button on the project page means "fix all stale envs for this app," and if you want per-env control, open the drawer.

Option 2 is probably fine for most use cases. The drawer already has per-env redeploy on both VariablesTab and DeploymentsTab. This issue is here to track the decision.

## Where the code lives

- `ui/src/routes/projects/[project]/+page.svelte` — `redeployOne()` function (line ~156) and stale-apps banner (line ~298)
- `internal/api/rebuild.go` — `RedeployStale` handler
- Drawer per-env redeploy: `ui/src/lib/components/drawer/VariablesTab.svelte:54`, `DeploymentsTab.svelte:40`

MC-Meesh:
## Summary
- Adds per-environment stale badges to both canvas and list views on the project page
- Each stale env badge is individually clickable, calling `api.redeploy(project, app, env)` to restart only that environment
- When an app has multiple stale envs, an "all" button reuses the existing `redeployStale` endpoint
- "Redeploy all" button still appears when multiple apps are stale
- Exports new `staleEnvNames()` utility from `types.ts`

Closes #213

## Test plan
- [ ] Verify stale env badges appear in canvas view floating banner per app
- [ ] Verify stale env badges appear inline in list view status column
- [ ] Click a single env badge and confirm only that env redeploys
- [ ] Click "all" button on an app with 2+ stale envs and confirm all restart
- [ ] Click "Redeploy all" when multiple apps are stale
- [ ] Confirm spinner states show correctly during redeploy
- [ ] Verify no regressions when no envs are stale (banner hidden)

### DIFF
--- a/ui/src/routes/projects/[project]/+page.svelte
+++ b/ui/src/routes/projects/[project]/+page.svelte
@@ -152,34 +152,58 @@
 		project?.autoRedeploy ? [] : apps.filter(a => appNeedsRedeploy(a))
 	);
 
+	/** Memoized stale env names per app — avoids redundant recomputation per render. */
+	const staleEnvMap = $derived(
+		new Map(apps.map(a => [a.metadata.name, staleEnvironments(a)]))
+	);
+
+	let redeployingApps = $state<Set<string>>(new Set());
 	let redeployingEnvs = $state<Set<string>>(new Set());
 	let redeployAllRunning = $state(false);
+	let redeployErrors = $state<Map<string, string>>(new Map());
 
-	function isAppRedeploying(appName: string) {
-		return [...redeployingEnvs].some(key => key.startsWith(appName + ':'));
+	const hasRedeployErrors = $derived(redeployErrors.size > 0);
+
+	async function redeployOne(appName: string) {
+		redeployingApps = new Set([...redeployingApps, appName]);
+		const errMap = new Map(redeployErrors);
+		errMap.delete(appName);
+		redeployErrors = errMap;
+		try {
+			await api.redeployStale(projectName, appName);
+		} catch (e) {
+			const msg = e instanceof Error ? e.message : `Redeploy failed for ${appName}`;
+			redeployErrors = new Map(redeployErrors).set(appName, msg);
+		} finally {
+			const next = new Set(redeployingApps);
+			next.delete(appName);
+			redeployingApps = next;
+		}
 	}
 
-	async function redeployEnv(appName: string, envName: string) {
+	/** Redeploy a single environment for one app. */
+	async function redeployOneEnv(appName: string, envName: string) {
 		const key = `${appName}:${envName}`;
 		redeployingEnvs = new Set([...redeployingEnvs, key]);
+		const errMap = new Map(redeployErrors);
+		errMap.delete(key);
+		redeployErrors = errMap;
 		try {
 			await api.redeploy(projectName, appName, envName);
-		} catch { /* error surfaces via SSE phase update */ }
-		finally {
+		} catch (e) {
+			const msg = e instanceof Error ? e.message : `Redeploy failed for ${appName}/${envName}`;
+			redeployErrors = new Map(redeployErrors).set(key, msg);
+		} finally {
 			const next = new Set(redeployingEnvs);
 			next.delete(key);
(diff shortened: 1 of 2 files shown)


MC-Meesh (commented):
**Code quality audit — PR #274**

**[WARNING]** ui/src/routes/projects/[project]/+page.svelte:redeployOneEnv — The `catch` block in `redeployOneEnv` silently swallows all errors (`catch { /* error surfaces via SSE phase update */ }`). If the API call fails due to a network error or auth issue (not just a deployment failure), the user sees no feedback — the button just stops spinning. SSE phase updates only fire for deployment-lifecycle events, not for API call failures. Consider at minimum logging the error or showing a toast for non-2xx responses.

**[NOTE]** ui/src/routes/projects/[project]/+page.svelte:redeployAllStale — `redeployAllStale` calls `redeployOne` (whole-app redeploy) for each stale app, but `redeployOneEnv` exists now for per-env granularity. If a user clicks "Redeploy all" they get whole-app redeploys even when only specific envs are stale. This is functionally correct but may cause unnecessary restarts for non-stale environments within a multi-env app. Worth a follow-up to wire `redeployAllStale` through `redeployOneEnv` for each stale env instead.

**[NOTE]** ui/src/lib/types.ts:staleEnvNames — `staleEnvNames` is called in multiple render locations (stale banner AND the table row for each app). For large project pages this means the filter+map runs redundantly on every render cycle. Not a real performance concern at current scale, but a `$derived` binding would be cleaner.

MC-Meesh (commented):

**Findings:**

### WARNING

1. **`staleEnvNames()` calls on every render in the list view:** In the list view table, `staleEnvNames(app)` is called inside the `{#each apps}` loop body to render per-env redeploy buttons. The function iterates `app.status?.environments` and filters/maps on every render cycle. For the typical case (a handful of apps, a few envs each) this is negligible. However, the same function is also called in the canvas banner `{#each staleApps}` block, meaning for stale apps it runs twice per render. Consider memoizing or computing once and passing as a derived value if performance becomes an issue with many apps.

2. **`redeployError` is shared across all apps/envs:** There is a single `redeployError` state variable for the entire page. If two concurrent redeploy operations fail (e.g., the user clicks two env badges quickly), the second error overwrites the first. Each `redeployOneEnv` and `redeployOne` call clears `redeployError` at the top before making the request, so a fast sequence of clicks would clear the error from a just-failed request before the user reads it. Consider a per-app or per-key error map instead.

### NOTE

1. **"Redeploy all" button hidden for single-app stale lists:** The `{#if staleApps.length > 1}` guard on the "Redeploy all" button means if only one app is stale (but it has multiple stale envs), the global "Redeploy all" button is hidden. The per-app "all" button still appears in that case (gated on `envs.length > 1`), so there is no functional gap. This is a reasonable UX choice.

2. **No E2E test coverage:** The PR description's test plan is entirely manual verification. Given the CLAUDE.md rule that "every button, link, input, toggle, and interactive element that exists in the UI must be exercised by at least one test," the new per-env redeploy buttons and error banner should have Playwright coverage. This may be intentional to add in a follow-up, but flagging it.

3. **`e.stopPropagation()` in list view buttons is correct:** The per-env redeploy buttons in the table rows correctly call `e.stopPropagation()` to prevent the row click handler from navigating away. Good attention to detail.

PR #275, opened 2026-05-08
fix: Helm chart security -- securityContext, RBAC docs, least privilege (#226)

### LINKED ISSUE #226: Helm chart security: runs as root with cluster-admin-equivalent RBAC
## Summary

The operator Deployment runs as root with no security context, and the ClusterRole grants cluster-wide access to all Secrets, all namespaces, and pods/exec in every namespace.

## Details

**P0-15: No securityContext (deployment.yaml:1-77)**

The Deployment template has no `securityContext` at either pod or container level. No `runAsNonRoot`, `readOnlyRootFilesystem`, `allowPrivilegeEscalation: false`, or `capabilities.drop: [ALL]`. The container runs as root.

**P0-16: Cluster-wide Secret CRUD (rbac.yaml:29-30)**

The ClusterRole grants full CRUD on all Secrets cluster-wide. A compromise of the operator gives access to every Secret in the cluster — TLS certs, other operators' credentials, service account tokens, etc.

**P0-17: Cluster-wide namespace CRUD (rbac.yaml:21-23)**

Full CRUD on all namespaces. Mortise only manages `pj-*` namespaces but the operator can delete `kube-system`.

**P0-18: Cluster-wide pods/exec (rbac.yaml:50-52)**

The operator can exec into any pod in any namespace, not just Mortise-managed workloads. Combined with Secret access, this is near-admin-level cluster access.

## Impact

A container escape gives the attacker root on the node + cluster-admin-equivalent RBAC = complete cluster compromise.

## Discussion needed

**securityContext** is straightforward to add. The RBAC discussion is harder:

Kubernetes RBAC cannot filter by resource name pattern. Options:
1. **Scoped RoleBindings per pj-* namespace** — the operator dynamically creates RoleBindings in each project namespace. More secure but adds reconcile complexity and a bootstrapping problem (needs namespace-create ClusterRole but scoped Roles for everything else).
2. **OPA/Kyverno policy** — admission-level restriction on the operator's SA to only touch `pj-*` namespaces. Adds an external dependency.
3. **Separate least-privilege SAs** — one ClusterRole for namespace/CRD operations, a second scoped Role for in-namespace resources. Reduces blast radius but doesn't eliminate it.
4. **Accept the risk with documentation** — document that the operator SA is privileged and should be treated as a cluster-admin equivalent. Recommend network policies to isolate the operator pod.

What's the right tradeoff here?

MC-Meesh:
## Summary

- **P0-15 (securityContext):** Add pod-level (`seccompProfile: RuntimeDefault`) and container-level (`allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: false`, `capabilities.drop: [ALL]`) security contexts to the operator Deployment. Add writable `/tmp` via emptyDir volume (64Mi limit) for future `readOnlyRootFilesystem` enablement.
- **P0-16/17/18 (RBAC scoping):** Break the monolithic `services, secrets, persistentvolumeclaims, serviceaccounts` rule into separate per-resource rules, each with a clear comment explaining WHY that permission is needed. Add a file-level RBAC SCOPING NOTES block documenting why cluster-wide access is still required for Secrets, Namespaces, and pods/exec (K8s RBAC cannot filter by namespace name pattern), and recommending OPA/Kyverno admission policies as mitigation.
- **Helm values:** Add `podSecurityContext` and `containerSecurityContext` values with hardened defaults, overridable per-environment.

> **Note:** `readOnlyRootFilesystem` is set to `false` and `runAsNonRoot` is omitted pending operator image hardening. The operator currently requires writable disk access beyond `/tmp`. These will be re-enabled once the operator's write paths are audited and confirmed compatible.

## Test plan

- [ ] `helm lint charts/mortise-core` passes (verified locally)
- [ ] `helm template` renders all resources correctly with security contexts applied (verified locally)
- [ ] Deploy to dev cluster and verify operator pod starts
- [ ] Verify operator can still reconcile Projects (namespace CRUD), Apps (deployments, services, secrets), and exec into pods
- [ ] Test overriding `podSecurityContext` / `containerSecurityContext` via values file

Closes #226

### DIFF
--- a/charts/mortise-core/templates/deployment.yaml
+++ b/charts/mortise-core/templates/deployment.yaml
@@ -17,10 +17,14 @@ spec:
         app.kubernetes.io/name: mortise
     spec:
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
         - name: mortise
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
+            {{- toYaml .Values.containerSecurityContext | nindent 12 }}
           ports:
             - name: api
               containerPort: {{ .Values.api.port }}
@@ -34,8 +38,14 @@ spec:
             - name: MORTISE_INGRESS_CLASS
               value: {{ .Values.operator.ingressClassName | quote }}
             {{- end }}
+            # Redirect temp writes to the writable emptyDir volume
+            - name: TMPDIR
+              value: /tmp
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+            - name: tmp
+              mountPath: /tmp
           livenessProbe:
             httpGet:
               path: /healthz
@@ -48,6 +58,10 @@ spec:
               port: 8081
             initialDelaySeconds: 5
             periodSeconds: 10
+      volumes:
+        - name: tmp
+          emptyDir:
+            sizeLimit: 64Mi
 ---
 {{- if .Values.ingress.enabled }}
 apiVersion: networking.k8s.io/v1
--- a/charts/mortise-core/templates/rbac.yaml
+++ b/charts/mortise-core/templates/rbac.yaml
@@ -1,52 +1,134 @@
+# -----------------------------------------------------------------------------
+# ClusterRole: mortise-controller
+#
+# This role grants the minimum permissions the Mortise operator needs.
+# Where possible, permissions are scoped by apiGroup, resource, and verb.
+#
+# RBAC SCOPING NOTES (see #226):
+#
+#   Secrets  — The operator creates per-App Secrets (image pull, env) in
+#              dynamically-created pj-* namespaces AND reads its own GitHub
+#              OAuth secret in mortise-system.  K8s RBAC cannot filter by
+#              namespace name pattern, so cluster-wide access is still
+#              required.  RISK: a bug in the operator could read/write
+#              Secrets in any namespace.  Mitigate with OPA/Kyverno
+#              policies that restrict Secret access to pj-* + mortise-system.
+#
+#   Namespaces — The Project controller creates/deletes one namespace per
+#              Project CRD.  Namespace names are dynamic (pj-{project}), so
+#              resourceNames cannot be used.  Cluster-wide is required.
+#
+#   Pods/exec — Used by the /api/projects/{project}/apps/{app}/exec
+#              endpoint to exec into user pods.  These pods only exist in
+#              pj-* namespaces.  Cluster-wide is granted because K8s RBAC
+#              cannot scope to a namespace name pattern.  Mitigate with
+#              OPA/Kyverno policies restricting exec to pj-* namespaces.
+# -----------------------------------------------------------------------------
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: mortise-controller
 rules:
-  # Project CRD (cluster-scoped; operator creates/deletes the backing namespace)
+  # ---------------------------------------------------------------------------
+  # Mortise CRDs — full lifecycle management
+  # The operator is the sole controller for all Mortise custom resources.
+  # ---------------------------------------------------------------------------
+
+  # Project CRD: cluster-scoped; operator creates/deletes the backing namespace.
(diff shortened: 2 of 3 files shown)


MC-Meesh (commented):
**Code quality audit — PR #275**

**[WARNING]** charts/mortise-core/values.yaml — `podSecurityContext` does not set `runAsUser` or `runAsGroup`. With only `runAsNonRoot: true`, the container image must define a non-root USER in its Dockerfile, otherwise the pod will fail to start with a `RunAsNonRoot` error. If the mortise image does not already set a non-root user, this change will break deployments. Suggest adding explicit `runAsUser: 1000` and `runAsGroup: 1000` (or whatever UID the image uses) to make the chart self-contained rather than depending on image metadata.

**[WARNING]** charts/mortise-core/templates/deployment.yaml — The `readOnlyRootFilesystem: true` + TMPDIR=/tmp + emptyDir approach is solid, but the operator may write to other paths besides /tmp (e.g., Go's default cache directory at `~/.cache`, or the working directory). If the operator binary writes anywhere outside `/tmp`, it will get `EROFS` at runtime. The integration test failures you mentioned may be related to this. Verify that the operator only uses TMPDIR for scratch writes, or add additional emptyDir mounts for known write paths (e.g., `/.cache`).

**[NOTE]** charts/mortise-core/templates/rbac.yaml — The RBAC refactor splits the old combined rule (`services, secrets, persistentvolumeclaims, serviceaccounts`) into individual rules. This is functionally equivalent and the documentation is excellent. The scoping notes acknowledging cluster-wide access for Secrets and Pods/exec with mitigation guidance (OPA/Kyverno) are a good pattern. No issues here.

MC-Meesh (commented):

**Findings:**

### WARNING

1. **`readOnlyRootFilesystem: false` in values.yaml contradicts the PR description:** The PR body says "Add container-level `readOnlyRootFilesystem: true`" but `values.yaml` sets `readOnlyRootFilesystem: false`. The emptyDir `/tmp` volume mount was specifically added to support a read-only root filesystem, but the default leaves it writable. This should be `true` to match the stated security posture. With `false`, the `/tmp` emptyDir still works but the security benefit of a read-only root is not realized.

2. **Missing `runAsNonRoot: true` in podSecurityContext:** The PR description says "Add pod-level (`runAsNonRoot`, `seccompProfile: RuntimeDefault`)" but the actual `podSecurityContext` in values.yaml only includes `seccompProfile`. The `runAsNonRoot: true` field is missing. Without it, the operator container could still run as root depending on the image's USER directive. This is a gap between the stated hardening and the implementation.

### NOTE

1. **RBAC comments are thorough and well-written:** The detailed scoping notes at the top of `rbac.yaml` explaining why Secrets, Namespaces, and Pods/exec require cluster-wide access (with OPA/Kyverno mitigation recommendations) are excellent documentation. This is the right way to handle the "we can't scope it but we know the risk" situation.

2. **Per-resource RBAC split is functionally equivalent:** Breaking the monolithic `services, secrets, persistentvolumeclaims, serviceaccounts` rule into individual rules with comments adds clarity without changing actual permissions. The verbs are identical across all four resources (`get, list, watch, create, update, patch, delete`), so there is no least-privilege improvement yet — but the separation makes future verb-scoping easier (e.g., if ServiceAccounts only need `get, create, delete`).

3. **Test fixture annotations (`mortise.dev/disable-default-security-context`):** These annotations appear in PR #272 as well, creating a cross-PR dependency. The annotation name implies there is code in the App controller that reads it and skips applying a default security context. That code is not visible in this diff — verify it exists on the target branch, or these annotations are inert.

MC-Meesh:
Follow-up audit on 2026-05-08:

The current PR head (`2247b29`) is still stale against `main` and GitHub is correctly marking it unmergeable. I verified the conflict locally with `git merge-tree`; the concrete collisions are in `charts/mortise-core/templates/deployment.yaml` and `charts/mortise-core/values.yaml`.

Two actionable points before this PR can be reviewed further:

1. Rebase or merge `main` into the branch and resolve the chart conflicts in favor of the current hardened shape already on `main`. The live PR head still references `containerSecurityContext`, drops the `runAsNonRoot` / UID / GID pod defaults, and sets `readOnlyRootFilesystem: false`, while current `main` already has hardened `podSecurityContext` + `securityContext` defaults.
2. After conflict resolution, rerun the Helm/chart checks on the updated branch. The RBAC documentation changes still look reasonable, but the security-context portion of the PR is no longer reviewable in isolation until the branch is brought up to date.

I did not see any unresolved GitHub review threads; the remaining blocker is the stale/conflicting branch state.

PR #297, opened 2026-05-10
[codex] fix env secret and preview cleanup edge cases

JamesSprow:
## Summary

This fixes a set of env/secret/preview lifecycle edge cases that were still reproducible after the BuildRun batch.

## What changed

- require secret API requests to resolve to a real `project + app + environment` tuple
- hide the internal `{app}-env` secret from user listings and block deleting it through the API
- preserve custom per-app secrets during project environment rename and clone flows
- precreate and then finalize project env namespaces to avoid races with project GC during env create/rename/clone
- reject deleting a project environment while app overrides still reference it
- delete app previews before dropping the app finalizer, and self-delete orphan previews when the parent app is gone
- skip deleting previews from the previews API once they are already terminating
- clean up preview env secrets correctly, while preserving `shared-env` until the last sibling preview for the same PR is gone

## Root causes

- several API handlers assumed the app/environment target existed and would happily operate on derived namespaces
- env rename/clone only carried CRD overrides, not the user-managed Secret state written through the API/UI
- app delete and preview GC depended on eventual cleanup paths that left stale preview resources and secrets behind
- preview secret GC treated `shared-env` like an app-owned secret even though it is shared per preview namespace/PR

## Validation

- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" go test ./internal/api -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" go test ./internal/controller -count=1`

## Notes

This branch also includes regression coverage for the API/controller paths above, including the shared preview secret cleanup case.


JamesSprow (commented):
REQUEST_CHANGES

I reviewed the PR diff plus the touched API/controller flows around env storage, app deletion, preview cleanup, and the relevant docs/spec.

Two blocking issues remain.

1. Secret-backed env vars are being written back into `App.Spec`
   - Files: `internal/api/project_envs.go`, adjacent contract in `internal/api/env.go`
   - `cloneEnvToApp` and `renameEnvOnApp` read values from the source `{app}-env` Secret and merge them into `App.Spec.Environments[].Env`.
   - That breaks the existing env-store model: the Secret is supposed to be the source of truth, and user-managed env vars are supposed to stay secret-backed rather than being copied into the App CRD.
   - Consequences:
     - secret values now become visible through normal App reads
     - secret state is duplicated into control-namespace CRD storage
     - the new tests/integration coverage lock this behavior in intentionally
   - I think the fix needs to preserve env state across rename/clone without syncing secret-backed values into `App.Spec`.

2. `DeleteApp` eagerly deletes secrets before the App delete is accepted
   - File: `internal/api/apps.go`
   - The handler deletes all Mortise-managed app secrets across env namespaces before calling `s.client.Delete(&app)`.
   - If the App delete fails, the request returns an error but the app still exists and its secrets are already gone.
   - That is a real correctness regression for live apps, and it duplicates cleanup that the app finalizer already owns in `internal/controller/app_env_gc.go`.
   - This should move back to finalizer-driven cleanup, or at least happen only after deletion is actually in progress.

Non-blocking but real doc drift:
- The stricter secrets API behavior in `internal/api/secrets.go` is not fully reflected in the checked-in OpenAPI/Swagger docs.
- Missing app / undeclared env / reserved internal secret delete now have tighter response behavior than the current annotations/spec describe.

Validation:
- Focused local envtest/controller runs passed:
  - `go test ./internal/api ./internal/controller -run 'Test(DeleteProjectEnvironmentRejectsReferencedOverride|UpdateProjectEnvironmentRenamePreservesSecretEnvVars|CloneProjectEnvironmentCopiesCustomSecrets|SecretEndpointsRequireExistingApp)|TestDeleteSecretRejectsInternalEnvSecret|TestDeleteAppCleansUpCustomSecretsViaAPI' -count=1`
  - `go test ./internal/controller -count=1`

Architecture/docs read:
- no third-party-SDK/controller rule violation
- no project/namespace model violation
- the main drift is the env-state contract, not the controller layering

JamesSprow:
Review summary: `REQUEST_CHANGES`

The preview cleanup direction and the stricter secret-target validation are both moving in the right direction, and I did not find a classic `CLAUDE.md` architecture violation such as controller SDK coupling or namespace-model drift.

The two blockers are:

1. Secret-backed env values are being copied into `App.Spec`
   - `internal/api/project_envs.go` now reads from `{app}-env` and writes those values into `App.Spec.Environments[].Env` during env rename/clone.
   - That conflicts with the existing env model where the Secret is the source of truth and user env vars remain secret-backed.
   - It also means values that were previously only in Secrets become visible through normal App CRD/API reads.

2. `DeleteApp` deletes secrets before the App delete is committed
   - `internal/api/apps.go` removes app-managed secrets before calling `Delete(app)`.
   - If the delete fails, the app remains but its secrets are already gone.
   - The app finalizer already owns cross-namespace cleanup, so this ordering is unsafe.

Docs/spec:
- I would not update spec/architecture docs to endorse the `App.Spec` secret-copy behavior; that should be fixed instead.
- If the stricter secret endpoint behavior stays, the OpenAPI/Swagger docs should be updated to describe the new `400`/`403`/`404` cases.

Test assessment:
- Focused envtest/controller runs passed locally.
- The missing coverage is on the actual failure modes above:
  - failed app delete must preserve secrets
  - rename/clone tests should stop asserting CRD backfill of secret-backed env values

JamesSprow:
Review scope: env secret / preview cleanup edges, with focus on environment rename/clone semantics, envstore ownership boundaries, and app deletion cleanup.

Findings:
- The clone/rename path was reading source env Secrets and merging those values back into `App.Spec.Environments[].Env`. That made envstore stop being the source of truth for secret-backed vars and leaked UI/API-set secret values through normal App reads.
- The same path could synthesize new env override entries from secret state alone, which is the wrong ownership boundary.
- `DeleteApp` was still eagerly deleting Mortise-managed app secrets in the API layer before the App delete committed.

Fixes pushed:
- Keep clone/rename CRD work limited to actual App override fields in `App.Spec`.
- Copy non-binding secret-backed env vars through envstore into the target environment namespace instead of copying them into `App.Spec`.
- Exclude binding-sourced vars from the copied Secret payload so they continue to be resolved by the controller in the target env.
- Remove API-side eager deletion of app-managed Secrets from `DeleteApp`; cleanup stays with controller/GC ownership.
- Add unit coverage proving rename/clone preserve secret-backed vars in envstore without leaking them into `App.Spec`, and update the integration contract accordingly.

Validation:
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -run 'Test(UpdateProjectEnvironmentRenamePreservesSecretEnvVars|CloneProjectEnvironment|DeleteApp)' -count=1`
- `make test-charts`
- `go test -v -parallel 1 -tags integration -count=1 -timeout 15m ./test/integration/... -run 'Test(RenameProjectEnvironmentPreservesSecretEnvVarsViaAPI|CloneProjectEnvironmentCopiesCustomSecretsViaAPI)$'`

JamesSprow:

PR #298, opened 2026-05-10
[codex] fix preview refresh after env updates

JamesSprow:
## Summary

This fixes stale preview environment secrets after app env updates.

## What changed

- add a filtered `App` watch to the `PreviewEnvironment` controller
- requeue existing previews for an app when the app spec changes or `mortise.dev/env-updated` changes
- keep the preview controller's existing secret rebuild path, but make sure it actually runs after env updates
- add controller coverage for refreshing preview secrets on a later reconcile
- add focused integration coverage for `PUT /api/projects/{project}/apps/{app}/env` refreshing the preview namespace `{app}-env` Secret

## Root cause

The env APIs already poke the `App` by updating `mortise.dev/env-updated`, but `PreviewEnvironmentReconciler` only watched preview-owned Deployments, Services, and Ingresses. Existing previews therefore never re-reconciled after app env changes, so preview namespace secrets stayed stale until some unrelated preview event happened.

## Validation

- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" go test ./internal/controller -count=1`
- `KUBECONFIG="$(k3d kubeconfig write mortise-int)" go test -tags integration -run TestPreviewEnvironmentRefreshesAfterEnvAPIUpdate -count=1 ./test/integration/...`

## Manual check

On `mortise-int` running `mortise:intfix4`, I reproduced the stale preview env bug against the live API, then verified the fix by:
- `PUT /api/projects/{project}/apps/{app}/env?env=production`
- confirming `mortise.dev/env-updated` changed on the `App`
- confirming the preview namespace `{app}-env` Secret updated from `LOG_LEVEL=info` to `LOG_LEVEL=debug`


JamesSprow (commented):
REQUEST_CHANGES

I reviewed the diff plus the adjacent preview/env-update flow in `internal/controller/previewenvironment_controller.go`, the preview controller tests, and the preview inheritance/spec docs.

There is one blocking correctness issue:

- The PR refreshes preview Secrets, but it still does not make the running preview workload pick up those new values.
- `reconcilePreviewEnvSecret` updates `{app}-env` / `shared-env`, and the same-SHA path now correctly continues into the Deployment reconcile path.
- But `reconcilePreviewDeployment` only references those Secrets via `EnvFrom`; it does not stamp a secret-derived pod-template annotation/hash like the main App controller does with `mortise.dev/env-hash`.
- That means a source env change can update the preview Secret objects without changing the preview pod template, so Kubernetes does not roll the preview pods. The live preview can keep running with stale env until some unrelated rebuild/redeploy happens.

Why this matters:
- The PR goal is to fix stale preview behavior after env updates.
- As written, it fixes stored Secret state, not the running preview environment.
- This also falls short of the existing preview inheritance contract described in `SPEC.md §5.8` and `docs/configuration.md`.

Tests are also not sufficient yet:
- The new refresh coverage checks Secret contents only.
- It does not assert a Deployment template mutation / rollout or that a running preview actually observes the new env after `PUT /api/projects/{project}/apps/{app}/env`.

What I’d like to see:
- Add a preview-side rollout trigger for env/secret changes, analogous to the main App controller’s env-hash pattern, or another equally explicit mechanism that forces the preview pods to restart when inherited env changes.
- Add regression coverage that proves the running preview picks up the updated env, not just that the Secret object changed.

Docs / architecture:
- I did not find a `CLAUDE.md` architecture violation.
- I do not think spec/architecture/general docs need changes for the intended fix once the runtime behavior is corrected.

JamesSprow:
Review summary: REQUEST_CHANGES

The new App watch is directionally right, and the PR does fix one real gap: existing previews now get requeued when the source App changes or `mortise.dev/env-updated` changes.

The blocking problem is that the change stops at Secret refresh. The preview controller rewrites the preview `{app}-env` / `shared-env` Secrets, but the preview Deployment path still has no env-hash-style pod-template trigger tied to that data. Since the pod template does not change, Kubernetes does not roll the preview pods, so a live preview can continue running with stale env even though the Secret objects were updated.

That means the PR currently fixes secret storage, not the running preview environment the user actually interacts with.

Test assessment:
- The added coverage checks that preview Secrets change on a later reconcile.
- It does not verify a preview rollout or that the running preview observes the new env after the API update.
- I also ran `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/controller -count=1`, which passed locally.

Docs/spec/architecture:
- No `CLAUDE.md` architecture issue found.
- No docs/spec update looks necessary for the intended fix.
- The problem is implementation correctness against the existing preview inheritance behavior already described in `SPEC.md` / `docs/configuration.md`.

Suggested direction:
- Add a preview-side env hash / rollout trigger, then add a regression test that proves the live preview picks up the updated values rather than only checking the Secret contents.

JamesSprow:
Review scope: preview refresh after env updates, specifically source-env watch/reconcile behavior, preview Secret refresh, and whether running preview pods actually pick up changed env.

Findings:
- The existing branch refreshed preview Secrets, but running preview Deployments could keep the old env because nothing forced a pod-template change.
- While validating live, I found a second issue: the preview controller wrote the refreshed Secret and then immediately recomputed `mortise.dev/env-hash` by rereading that Secret through the cached controller client. In a real cluster that same-reconcile read could be stale, so the Deployment kept the old hash and never rolled.

Fixes pushed:
- Reuse the env-secret hashing helper so preview Deployments carry the same `mortise.dev/env-hash` rollout trigger as normal app workloads.
- Compute the preview env hash from the just-merged desired env payload instead of rereading the cached Secret after writing it.
- Extend controller coverage to assert the preview Deployment hash changes after a later reconcile.
- Strengthen the integration test to prove the preview Secret updates, the Deployment template hash changes, the pod UID changes, and the live container env reflects the new value.

Validation:
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/controller -run TestControllers -count=1`
- `make test-charts`
- `go test -v -tags integration -count=1 -timeout 15m ./test/integration/... -run TestPreviewEnvironmentRefreshesAfterEnvAPIUpdate`

JamesSprow:

PR #299, opened 2026-05-10
[codex] fix exec app validation and pod selection

JamesSprow:
## Summary
- validate the project/app/environment tuple before exec pod lookup
- require a running, non-terminating app pod and target its first app container explicitly
- harden exec error handling and add regression coverage for missing-app and pod lookup failure paths

## Root cause
The exec handler trusted path parameters too early and selected pods too loosely. That let phantom apps target matching labeled pods, could choose non-running pods, and masked infrastructure lookup failures as ordinary 404s.

## Validation
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" go test ./internal/api -run 'Test(Exec|FindAppPod)' -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" go test ./internal/api -count=1`


JamesSprow (commented):
REQUEST_CHANGES

This is close, and the project/app/environment validation plus the pod lookup hardening are both good improvements. I found one blocking correctness issue in the new container-targeting path.

High:
- `internal/api/exec.go:157-170`
  The handler now passes an explicit container name into `pods/exec`, but `firstAppContainerName` still just returns `pod.Spec.Containers[0].Name`.

  That is not the same as “the app container.” Mortise’s own Deployment builder names the workload container `app.Name` (`internal/controller/app_controller.go:986-999`), so this code has enough information to select the intended container explicitly, but it does not use it.

  Why this matters: `CLAUDE.md` is explicit that Mortise must coexist with other Kubernetes tooling. Live Pods can be mutated by admission webhooks or service meshes that inject sidecars and may affect container ordering. In that case `/exec` can land in the sidecar instead of the workload container, which is exactly the kind of loose targeting this PR is trying to eliminate.

Medium:
- `internal/api/exec_internal_test.go`
  The new tests cover the missing-app path, pod lookup failures, terminating pods, and non-running pods, but they do not cover the multi-container / injected-sidecar case. Please add a regression test with a matching running Pod where container 0 is not the app container and assert that `/exec` still targets the workload container.

Docs / architecture:
- I did not find a required `SPEC.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, or `DEVELOPMENT.md` update for the intended fix.
- `internal/api/openapi.yaml` already has the `truncated` field on `execResponse`, so I do not see an API-doc drift blocker here.
- The PR otherwise stays within the repo’s architecture rules; the blocker is the container-selection detail.

Validation:
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -run 'Test(Exec|FindAppPod)' -count=1`
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -count=1`

Both passed on the PR head.

JamesSprow:
Review summary: REQUEST_CHANGES

The PR fixes real problems:
- it validates the project/app/environment tuple before pod lookup
- it stops missing apps from targeting labeled phantom pods
- it distinguishes pod lookup failures (`500`) from “no running pod” (`404`)
- it skips terminating and non-running pods

The remaining blocker is container selection inside the chosen Pod:
- `internal/api/exec.go` now sends an explicit container name to `pods/exec`, but it derives that name by taking `pod.Spec.Containers[0]`
- Mortise’s own reconciler names the workload container `app.Name`
- because Mortise is supposed to coexist with other Kubernetes tooling, a live Pod may have injected sidecars or reordered containers
- in that case `/exec` can still land in the wrong container

I do not think this PR needs spec/architecture/general doc updates once that is corrected. The OpenAPI already reflects the current `execResponse` shape.

Test assessment:
- focused exec tests passed on the PR head
- full `go test ./internal/api -count=1` also passed
- the missing regression case is a multi-container Pod where the app container is not first

Net: the request-routing and pod-selection hardening are good, but the implementation needs to target the actual app container rather than container index 0 before this is ready.

JamesSprow:
Reviewed the exec-path hardening in this PR and fixed one remaining correctness bug.

Findings:
- `ExecInApp` already routes through the app exec endpoint, but pod selection still picked `pod.Spec.Containers[0]`, so a sidecar-first pod could exec into the wrong container.
- The old swagger/openapi description still said the endpoint runs against the first running pod, which no longer matched the intended behavior.

Fixes pushed in `5edaa37`:
- select the container whose name matches the app name, with a single-container fallback only
- add a regression test covering a sidecar-first pod
- update the `/exec` swagger/openapi description to reflect explicit app-container targeting
- remove the dead `projectName` assignment that was tripping `staticcheck`

Local validation:
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -run 'Test(Exec|FindAppPod)' -count=1`
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -count=1`
- `go vet ./internal/api/...`
- `staticcheck ./internal/api/...`

GitHub Actions reran automatically on the new head SHA; I am treating that as the final CI gate for this PR.

PR #300, opened 2026-05-10
[codex] fix proxy environment resolution

JamesSprow:
## Summary
- adopt PR #300 as the vehicle for issue #314 and carry the branch onto current `main`
- extend defaulted app-environment validation across the remaining app endpoints: `pods`, `exec`, traffic history/current, `redeploy`, and app secret routes
- keep proxy `connect` validation aligned with the same `400` undeclared / `404` app-disabled model
- fix the reviewed proxy cleanup regression so `disconnect` can still remove an active stale listener after a project env is renamed or deleted
- regenerate the OpenAPI docs for the new `400` responses and documented proxy disconnect query parameter

## Root cause
PR #306 fixed declared-environment validation for logs and metrics, but several sibling endpoints still defaulted `env` and acted on derived namespaces without checking that:
- the env is declared on the project
- the app participates in that env

PR #300 had already started the proxy slice of that work, but it still needed the stale-disconnect cleanup fix and the API doc update. PR #299 overlapped on `exec`; this branch carries only the env-validation portion needed for issue #314, not the broader pod-selection work from that PR.

## Notes on the adopted vehicle
- The branch now includes current `main`.
- That rebase path pulls in the shared admin bootstrap helper fix from `main`, which is the narrow change relevant to the prior `TestGiteaOAuthFlow` failure on this PR vehicle.
- No extra integration-harness patch was required on top of `main` for the #314 scope.

## Validation
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -run 'Test(Exec|HandlePods|Traffic|HandleConnect|HandleDisconnect|Redeploy|Secrets|MemberCanCRUDSecrets)' -count=1`
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -count=1`
- `make generate-api`


JamesSprow (commented):
Requesting changes for one behavioral regression and one API-doc gap.

Findings:
- `internal/api/proxy.go:97-116,238-256`: `handleDisconnect` now validates the requested env against the current `Project.spec.environments` before it checks the in-memory proxy map. That breaks cleanup for stale listeners when a project environment is renamed or deleted after the proxy was started. In that case the request now returns `400` and the local listener remains alive until the API process restarts. Since project envs are mutable and proxy listeners are local process state, disconnect needs to remain able to tear down an existing listener keyed by the old env name.
- `internal/api/proxy.go:229-237` / `internal/api/openapi.yaml:3272-3298`: this PR changed `disconnect` to resolve/default an environment and to return `400` on invalid envs, but the generated API docs still expose only the path params and `204/404`. The public contract is incomplete.

What looked good:
- The core fix is aligned with the project/environment model: `connect` now validates against declared project envs, defaults to the first project env, and refuses proxies for app envs explicitly disabled with `enabled: false`.
- I did not find any `CLAUDE.md` architecture-rule violation or any required `SPEC.md` / `ARCHITECTURE.md` narrative change for the intended behavior.

Tests:
- The added proxy tests cover undeclared env rejection, disabled env rejection, first-env defaulting, and per-env scoping.
- Please add a regression test proving a proxy created for an env can still be disconnected after that env is removed or renamed on the project.

JamesSprow:
Review verdict: `REQUEST_CHANGES`

Primary issue:
- `disconnect` now validates the env against the current project before checking the active proxy map. That means if a proxy was started for `staging` and the project environment is later renamed or deleted, `/disconnect?env=staging` now fails with `400` and leaves the listener running. This is a cleanup regression in the touched project-env flow.

Documentation/API contract:
- `connect` already documents its env behavior, but `disconnect` still does not document the environment query parameter or the new `400` invalid-env response. The generated OpenAPI should be updated with this PR.

Docs/spec/architecture:
- No `CLAUDE.md` architecture drift found.
- No `SPEC.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, or `DEVELOPMENT.md` narrative changes look necessary beyond the API doc fix above.

Tests:
- The new tests are useful and cover the main happy-path/env-validation changes.
- They should be extended with a stale-proxy cleanup case: create a proxy for one env, mutate the project env list, then verify disconnect can still remove the existing listener.

JamesSprow:
Closing as superseded by #326. The retained branch carries the refreshed env-validation work plus the CLI proxy teardown env-selection fix, so there is no separate surviving merge scope here.

PR #302, opened 2026-05-10
[codex] fix deploy token cleanup on app delete

JamesSprow:
## Summary
- delete app-scoped deploy token secrets when an app is removed through the REST API
- extend App finalizer GC to remove app-scoped deploy token secrets for non-API deletion paths too
- add regression coverage that preserves project-scoped deploy tokens while removing app-scoped ones

## Root cause
App deploy tokens were stored as standalone Secrets in the project control namespace with no owner reference and labels that did not match the existing finalizer GC selector. Deleting an app left valid deploy credentials behind, which could leak into same-name app recreations.

## Validation
- `KUBEBUILDER_ASSETS="/tmp/mortise-pr-exec-fixes/bin/k8s/1.35.0-linux-amd64" go test ./internal/api ./internal/controller -count=1`

### DIFF
--- a/internal/api/apps.go
+++ b/internal/api/apps.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/go-chi/chi/v5"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -324,6 +325,11 @@ func (s *Server) DeleteApp(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := s.deleteAppDeployTokens(r.Context(), ns, app.Name); err != nil {
+		writeError(w, r, err)
+		return
+	}
+
 	if err := s.client.Delete(r.Context(), &app); err != nil {
 		writeError(w, r, err)
 		return
@@ -332,6 +338,25 @@ func (s *Server) DeleteApp(w http.ResponseWriter, r *http.Request) {
 	s.recordActivity(r, projectName, "delete", "app", app.Name, "Deleted app "+app.Name, "")
 
 	writeJSON(w, http.StatusOK, map[string]string{"status": "deleted"})
+}
+
+func (s *Server) deleteAppDeployTokens(ctx context.Context, ns, appName string) error {
+	var tokens corev1.SecretList
+	if err := s.client.List(ctx, &tokens,
+		client.InNamespace(ns),
+		client.MatchingLabels{
+			"mortise.dev/deploy-token": "true",
+			"mortise.dev/app":          appName,
+		},
+	); err != nil {
+		return err
+	}
+	for i := range tokens.Items {
+		if err := s.client.Delete(ctx, &tokens.Items[i]); err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	}
+	return nil
 }
 
 // writeError maps k8s API errors to HTTP status codes.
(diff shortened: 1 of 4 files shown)


JamesSprow (commented):
Review summary: request changes.

The cleanup goal is correct, and the controller-side finalizer update is the right direction. I found one blocking issue in the API delete path.

High:
- `internal/api/apps.go:320`
  `DeleteApp` removes app-scoped deploy token secrets before it submits the App delete to the API server. If `s.client.Delete(&app)` fails, the request returns an error but the app still exists and all deploy tokens are already revoked. That makes the endpoint non-atomic and can break CI for a live app after a failed delete attempt.

  Now that `gcAppAcrossEnvs()` also cleans app-scoped deploy token secrets during finalizer GC, the API path should not eagerly revoke them before the App delete is accepted. Either:
  - rely on the finalizer cleanup only, or
  - move token cleanup to a point where a failed App delete cannot strand a live app without credentials.

Medium:
- `internal/api/server_test.go:494`
  The new regression coverage only exercises the success path. Please add a test that forces the App delete to fail and proves app-scoped deploy tokens remain intact in that case.

Docs / architecture:
- I do not think this bugfix requires `SPEC.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, or `DEVELOPMENT.md` updates.
- I did not find a `CLAUDE.md` architecture-rule violation in the touched code.

Tests:
- The added envtest coverage is useful, especially the project-token preservation assertions.
- It still needs failure-path coverage for the API delete ordering described above.

JamesSprow:
Review summary: request changes.

The finalizer-side cleanup in `internal/controller/app_env_gc.go` looks like the right long-term fix for non-API deletion paths, and the new tests correctly protect project-scoped tokens from being deleted.

The blocking issue is in `internal/api/apps.go`: `DeleteApp` deletes app-scoped deploy token secrets before the App delete is submitted. If the App delete fails, the request errors out but the app still exists and its deploy tokens are already gone. That creates a real regression risk for live apps and CI.

I don’t think this PR needs spec/architecture/general doc updates once that behavior is corrected, because it does not change the public API contract or product model. Test coverage is close, but it should also include a failed-delete case proving tokens are preserved when the App delete does not succeed.

JamesSprow:
Review scope: deploy token lifecycle on app deletion, especially failure ordering and app-vs-project token ownership.

Findings:
- `DeleteApp` revoked app-scoped deploy token secrets before deleting the App resource. If the app delete failed, the app stayed live but lost its deploy credentials.

Fixes pushed:
- Reorder `DeleteApp` so the App resource is deleted first and app-scoped deploy token cleanup happens only afterward.
- Add a regression test proving deploy token secrets are preserved when app deletion fails.

Validation:
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -run 'TestDeleteApp|TestDeleteAppRemovesAppDeployTokensOnly|TestDeleteAppPreservesDeployTokensWhenAppDeleteFails' -count=1`
- `KUBEBUILDER_ASSETS=/home/james/workspaces/mortise/bin/k8s/1.35.0-linux-amd64 go test ./internal/controller -run 'TestControllers|TestPreviewSyncStateIncludesBackfillInputs' -count=1`
- `make test-charts`

JamesSprow:

JamesSprow:
Refreshed onto current `main` and reran the delete-ordering pass.

Local validation passed:
- `go test ./internal/api -run 'TestDeleteApp|TestDeleteAppRemovesAppDeployTokensOnly|TestDeleteAppPreservesDeployTokensWhenAppDeleteFails' -count=1`\n- `go test ./internal/controller -run 'TestControllers|TestPreviewSyncStateIncludesBackfillInputs' -count=1`\n- `go test ./internal/api -count=1`\n- `go test ./internal/controller -count=1`\n\n`DeleteApp` still deletes the App before any app-scoped deploy-token cleanup, and finalizer GC remains responsible for the cross-env/app-owned token cleanup path. Fresh CI is now running on `5399bf2`.

PR #303, opened 2026-05-10
[codex] fix proxy connect listener race

JamesSprow:
## Summary
- serialize same-process proxy connect creation so one key only creates one listener
- add a deterministic concurrency regression using a blocked listener seam
- ensure concurrent same-key connects return the same proxy URL instead of leaking an orphan listener

## Root cause
`handleConnect` checked for an existing proxy under lock, dropped the lock, then performed app lookup and `Listen` before storing the entry. Two concurrent requests for the same key could both create listeners and only the last one stayed tracked.

## Validation
- `KUBEBUILDER_ASSETS="/tmp/mortise-pr-exec-fixes/bin/k8s/1.35.0-linux-amd64" go test ./internal/api -run 'Test(HandleConnectCreatesOneListenerPerKey|Proxy)' -count=1`

### DIFF
--- a/internal/api/proxy.go
+++ b/internal/api/proxy.go
@@ -21,6 +21,8 @@ import (
 	"github.com/mortise-org/mortise/internal/constants"
 )
 
+var proxyListen = net.Listen
+
 type retryTransport struct {
 	base     http.RoundTripper
 	retries  int
@@ -127,18 +129,17 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 	envNs := constants.EnvNamespace(projectName, env)
 	key := proxyKey(projectName, appName, env)
 
-	// If already proxying, return the existing URL.
 	s.proxies.mu.Lock()
 	if entry, exists := s.proxies.proxies[key]; exists {
 		s.proxies.mu.Unlock()
 		writeJSON(w, http.StatusOK, entry)
 		return
 	}
-	s.proxies.mu.Unlock()
 
 	// Resolve app CRD (control ns) to get the port.
 	var app mortisev1alpha1.App
 	if err := s.client.Get(r.Context(), types.NamespacedName{Name: appName, Namespace: ns}, &app); err != nil {
+		s.proxies.mu.Unlock()
 		writeJSON(w, http.StatusNotFound, errorResponse{"app not found"})
 		return
 	}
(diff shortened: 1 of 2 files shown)


JamesSprow:
Review verdict: `REQUEST_CHANGES`

Primary issue:
- The race fix in `internal/api/proxy.go` is functionally too broad. `handleConnect` now keeps the global proxy mutex locked from the initial cache miss all the way through app lookup and listener allocation. That prevents duplicate listeners for the same key, but it also serializes unrelated proxy operations across the whole server. One slow `connect` can now stall other apps/environments and block `disconnect` as well.

Docs/spec/architecture:
- I did not find a `CLAUDE.md`, `SPEC.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, or `DEVELOPMENT.md` mismatch that requires documentation changes for the intended fix.
- The concern is implementation behavior, not documented product shape.

Tests:
- The new deterministic same-key race test is valuable and directly targets the original bug.
- It should be paired with a different-key concurrency test so we prove the fix is scoped to one proxy key rather than turning the proxy manager into a process-wide bottleneck.

Suggested direction:
- Keep deduplication per proxy key, but avoid holding the one global mutex across Kubernetes reads and `Listen`.

JamesSprow (commented):
This fixes the duplicate-listener race for the same proxy key, but I’m requesting changes because the locking scope is now too broad.

Finding:
- `internal/api/proxy.go:132-195` holds the single global `s.proxies.mu` across `s.client.Get(...)`, `url.Parse(...)`, and `proxyListen(...)`. That means a slow/hung connect for one app/environment now blocks all other `connect` requests, and it also blocks `disconnect` because `handleDisconnect` takes the same mutex. The PR summary says the goal is to ensure one key only creates one listener, but the current implementation serializes proxy lifecycle operations process-wide rather than per key.

What I’d like to see:
- Narrow the serialization to the specific proxy key being created, for example with a per-key pending entry or similar singleflight-style coordination.
- Add a regression test proving that two different keys can still connect independently while same-key requests are deduplicated.

Everything else I checked looks fine from a doctrine/docs standpoint:
- no `CLAUDE.md` architecture-rule violation
- no spec/architecture/doc update required for the intended fix
- the same-key regression test is useful, but it needs the different-key companion coverage

MC-Meesh:
Addressed the locking-scope review on new head `b415aec`.

What changed:
- Replaced the broad global connect critical section with per-proxy-key pending coordination in `internal/api/proxy.go`.
- Same-key callers now deduplicate by waiting on the in-flight key and then re-reading the finished entry.
- Different keys no longer serialize behind one slow `connect`; the global mutex only protects the shared maps.
- No docs update appears needed because this restores the intended per-key behavior rather than changing product semantics.

Validation:
- `KUBEBUILDER_ASSETS=/Users/meesh/capybara/bin/k8s/1.35.0-darwin-arm64 go test ./internal/api -run "TestHandleConnect(CreatesOneListenerPerKey|DoesNotSerializeDifferentKeys)" -count=1`
- `KUBEBUILDER_ASSETS=/Users/meesh/capybara/bin/k8s/1.35.0-darwin-arm64 go test ./internal/api -run "Test(HandleConnectCreatesOneListenerPerKey|HandleConnectDoesNotSerializeDifferentKeys|Proxy)" -count=1`

The new regression `TestHandleConnectDoesNotSerializeDifferentKeys` blocks the first key inside `proxyListen`, then proves a different app/env key still reaches listener allocation and completes independently.

MC-Meesh:
@JamesSprow requested changes are addressed on green head `b415aec`.

Explicit response to the review:
- The global proxy mutex is no longer held across app lookup, target parsing, or `proxyListen`.
- Same-key dedupe is preserved with per-key pending coordination only.
- Different keys now proceed independently again; the global lock only protects the shared proxy maps.
- Added the companion regression `TestHandleConnectDoesNotSerializeDifferentKeys` to prove unrelated keys are not globally serialized.
- No docs update appears required because this restores the intended per-key behavior rather than changing documented semantics.

Fresh CI run `25648954568` is fully green on this head.

PR #304, opened 2026-05-10
[codex] fix deploy environment targeting and auth

JamesSprow:
## Summary
- validate deploy target environments against the project before mutating app state
- write env-targeted deploys to `spec.environments[].image` and synthesize the override when the app participates implicitly
- enforce restricted-env authz for both explicit env deploys and global deploys that would touch restricted environments

## Root cause
The deploy handler had drifted from the app/env model. It validated authz with optional environment context, but it still treated missing app overrides as missing environments and allowed global deploys to bypass restricted-env checks by omitting `environment`.

## Validation
- `KUBEBUILDER_ASSETS="/tmp/mortise-pr-exec-fixes/bin/k8s/1.35.0-linux-amd64" go test ./internal/api -run 'TestDeploy' -count=1`

### DIFF
--- a/internal/api/deploy.go
+++ b/internal/api/deploy.go
@@ -96,29 +96,46 @@ func (s *Server) Deploy(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	var app mortisev1alpha1.App
-	if err := s.client.Get(r.Context(), types.NamespacedName{Name: appName, Namespace: ns}, &app); err != nil {
-		writeError(w, r, err)
+	project, ok := s.getProject(w, r)
+	if !ok {
 		return
 	}
-
 	if req.Environment != "" {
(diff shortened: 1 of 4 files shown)


MC-Meesh:
Staff review on final green head `d365a10`: deploy environment targeting changes remain clean, and the added `LoginAsAdmin` bounded retry is the correct fix for the remaining parallel setup/login race seen in integration. Local validation passed on `go test ./test/helpers -count=1`, `go test ./test/integration -tags integration -run "TestGitProviderAdminAPICRUD" -count=1`, `KUBEBUILDER_ASSETS=/Users/meesh/capybara/bin/k8s/1.35.0-darwin-arm64 go test ./internal/api -run "TestDeploy" -count=1`, and `go vet ./...`. No further findings on this head.

PR #308, opened 2026-05-10
[codex] fix gitprovider webhook secret reads

JamesSprow:
## Summary
- return `404` when a git provider references a webhook secret key that does not exist instead of returning `200` with an empty value
- preserve the existing `200` empty response only for the explicit `WebhookSecretRef == nil` case
- add API regression coverage for the broken referenced-key path

## Testing
- `KUBEBUILDER_ASSETS="/tmp/mortise-pr-exec-fixes/bin/k8s/1.35.0-linux-amd64" go test ./internal/api -run 'TestGetWebhookSecret(MissingReferencedKeyReturns404|ForbiddenForMember)' -count=1`

### DIFF
--- a/internal/api/gitproviders.go
+++ b/internal/api/gitproviders.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
@@ -263,7 +264,15 @@ func (s *Server) GetWebhookSecret(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	v := string(secret.Data[gp.Spec.WebhookSecretRef.Key])
+	raw, ok := secret.Data[gp.Spec.WebhookSecretRef.Key]
+	if !ok {
+		writeJSON(w, http.StatusNotFound, errorResponse{
+			fmt.Sprintf("secret %s/%s does not contain key %q", gp.Spec.WebhookSecretRef.Namespace, gp.Spec.WebhookSecretRef.Name, gp.Spec.WebhookSecretRef.Key),
+		})
+		return
+	}
+
+	v := string(raw)
 	writeJSON(w, http.StatusOK, map[string]string{"webhookSecret": v})
 }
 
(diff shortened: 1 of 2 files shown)


MC-Meesh (approved):
Reviewed internal/api/gitproviders.go and internal/api/gitproviders_test.go. The change correctly distinguishes an explicitly missing referenced secret key from the intentional nil-WebHookSecretRef case, preserves the empty-string response only for the nil-ref path, and adds focused regression coverage for the broken key lookup plus existing authorization behavior. Focused local envtest validation on the new cases passed.

JamesSprow (commented):
Reviewed `internal/api/gitproviders.go` and `internal/api/gitproviders_test.go`, plus the adjacent secret-ref handling paths in `internal/controller/helpers.go` and `internal/webhook/k8s.go`.

No blocking issues found.

This change fixes the real bug: a `GitProvider` that references a Secret key that does not exist now returns `404` instead of silently returning an empty webhook secret. That aligns this admin read endpoint with the existing secret-ref expectations elsewhere in the codebase: the controller already treats missing keys as invalid, and the webhook runtime path already errors when a referenced key is absent.

The nil `WebhookSecretRef` case remains intentionally distinct and still returns the empty-string payload, which matches the current endpoint contract.

Focused regression coverage for the missing-key case is in place, and the member-forbidden authz path remains covered. I did not find any required `SPEC.md`, `ARCHITECTURE.md`, or general docs updates for this change.

JamesSprow:
Verdict: approve.

I reviewed the diff in `internal/api/gitproviders.go` / `internal/api/gitproviders_test.go` and checked the adjacent touched flow around secret-ref validation and webhook secret consumption.

Findings:
- No blocking correctness or architecture issues.
- The handler now correctly distinguishes:
  - `WebhookSecretRef == nil` -> `200 {"webhookSecret": ""}`
  - referenced Secret key missing -> `404`
- That behavior is consistent with the rest of the codebase: controller-side secret-ref validation already rejects missing keys, and the webhook runtime loader already errors on missing keys.

Docs/spec/architecture:
- No `CLAUDE.md`, `SPEC.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, or `DEVELOPMENT.md` updates appear necessary.
- The existing OpenAPI already allows `404` for this endpoint, so this is a bug fix within the documented contract, not a surface change.

Tests:
- The new regression test covers the broken referenced-key path directly.
- The existing forbidden-for-member coverage still protects the authz side.
- Minor non-blocking gap: there is still no explicit test for the preserved nil-ref `200` empty response path.

Issue #295, opened 2026-05-09
webhook auto-registration retries forever on persistent GitHub 404 failures


JamesSprow:
## Summary
When automatic GitHub webhook registration fails with a persistent 404, Mortise retries on every reconcile and floods controller logs.

In a live cluster this showed up as repeated pairs of errors for the same apps:
- `failed to list webhooks, proceeding with registration`
- `webhook registration failed (non-fatal, builds still work manually)`

The underlying GitHub API responses were 404 on both list and register attempts.

## Why this is a Mortise bug
`ensureWebhook()` only short-circuits when the `mortise.dev/webhook-registered` annotation is set.
If webhook registration fails permanently because of repo visibility, permissions, or provider mismatch, the controller records nothing and just retries forever on subsequent reconciles.

That means a single bad repo-hook setup can produce endless noisy retries instead of a latched degraded condition.

## Code path
- `internal/controller/app_controller.go`
  - `ensureWebhook()`

Current behavior:
- tries `ListWebhooks()`
- logs and continues even on error
- tries `RegisterWebhook()`
- returns error to caller on failure
- no durable "do not retry until spec/token/provider changes" state is recorded

## Observed impact
On the cluster, apps against `https://github.com/hayden3456/wax_letter.git` produced repeated 404s for webhook list and register calls over and over within the same period.

Builds still worked manually, but the controller spammed logs and kept wasting reconcile work.

## Expected behavior
Webhook auto-registration failures that are effectively permanent should not be retried on every reconcile forever.

## Suggested fix
- Record a durable failure condition / annotation for webhook auto-registration attempts
- Retry only when relevant inputs change (repo URL, provider config, token owner, domain/webhook URL, explicit user action)
- Distinguish permission/not-found failures from transient network/server errors

Issue #290, opened 2026-05-09
operator restart drops in-memory build tracker and re-runs active git env builds from scratch


JamesSprow:
## Summary
In `v1.0.0`, active git builds are tracked only in-memory. If the operator restarts while an env build is in progress, the next reconcile re-launches that build from scratch for the same `app + env + revision`.

The current code comments say this is acceptable because builds are idempotent, but in practice it is expensive and user-visible:
- duplicated clone/build work
- canceled or overlapping BuildKit solves
- apps left showing `Building` until a later status pass wins

## Why this is not just migration fallout
This does not depend on old pre-`v1` status layout. It follows directly from the `v1.0.0` controller design:
- build tracker store is in-memory only
- restart loses the tracker map
- next reconcile sees no active tracker and can start the build again if env-level success markers were not yet persisted

We hit this immediately after upgrade because the operator itself was restarted/evicted, but the same behavior should happen on any normal controller restart.

## Code paths
- `internal/controller/build_tracker.go`
  - `buildTrackerStore` comment explicitly notes that operator restarts lose the map and the next reconcile will re-launch
- `internal/controller/app_controller.go`
  - `reconcileEnvBuild()` uses the in-memory tracker as the only running-build guard
  - if the tracker is missing and env-level `LastBuiltSHA` / `LastBuiltImage` are not yet persisted, the same env build is started again

## Expected behavior
A controller restart should not require redoing an already running build for the same env/revision unless there is no recoverable way to resume or detect completion.

## Suggested directions
- Persist in-progress build state outside process memory so a restarted controller can resume/poll instead of blindly relaunching.
- Or record a durable build token/status on the App so a restart can distinguish `running but not yet persisted` from `never built`.
- At minimum, document the user-visible impact more clearly if restart-driven rebuilds are considered an intentional tradeoff.

MC-Meesh:
Closing as **duplicate of #447**. Verified against HEAD `5e3011e` in the mo-4jv quality audit:

- The specific mechanism filed here is stale: builds are now guarded by durable `BuildRun` CRs, not the in-memory tracker (`internal/controller/app_controller.go:768-853` checks `CurrentBuildRunRef`/`LastBuiltSHA` against the API), and only `BuildRunReconciler` launches builds
[...]

Issue #281, opened 2026-05-08
Project environment creation race: staging POST can clobber default production env


MC-Meesh:
## Problem

When the UI creates a project and immediately POSTs a staging environment, a race condition can cause the default production environment to be lost:

1. UI creates project via POST /api/projects
2. Controller begins reconciling -- adds default "production" environment to spec
3. UI POSTs staging environment before the controller's update completes
4. The staging POST reads the project spec (without production yet), adds staging, and calls `client.Update`
5. The controller's concurrent update (adding production) and the API's update (adding staging) conflict
6. Whichever write wins, the other environment is lost

### Result
The project ends up with only `["staging"]` instead of `["production", "staging"]`.

### Suggested fix
The environment creation API should use a patch/merge strategy instead of a full spec replacement, or use `client.MergeFrom` to avoid clobbering concurrent changes.

Alternatively, the API could wait for the controller to add default environments before accepting new environment creation requests.

Found during #228.

MC-Meesh:
**Audit re-scope (mo-4jv, HEAD `5e3011e`):** the filed mechanism is fixed, but the symptom is still reachable via a different race.

- **Fixed:** `createProjectEnvironment` now wraps the read-modify-write in `retry.RetryOnConflict` with a fresh `Get` inside the closure and appends (`internal/api/project_envs.go:213-225`); the sibling env handlers are consistently conflict-safe. "Whichever write wins, the other environment is lost" no longer holds.
- **Still broken (same `["staging"]`-only outcome):** `CreateProject` creates the Project with no environments (`internal/api/projects.go:183-186`); the controller seeds `production` only when `len(Spec.Environments) == 0` (`internal/controller/project_controller.go:163-168`); the UI POSTs `staging` immediately after create with a swallowed error (`ui/src/routes/projects/new/+page.svelte:57-64`). If the staging append lands before the controller's seed pass, the seed branch is skipped forever — no conflict occurs, so the retry loop can't help.
- **Suggested fix:** seed `production` in `CreateProject` itself (closes the window), and change the controller guard from `len == 0` to "no non-preview environment present" as a backstop (the controller seed must stay for `kubectl`-created Projects). Size S.

Issue #280, opened 2026-05-08
SettingsTab envEnabled derives from prop instead of specOverride


MC-Meesh:
## Problem

In `SettingsTab.svelte` line 30:
```js
let envEnabled = $derived(envEntry?.enabled !== false);
```

`envEntry` comes from the `app` prop (passed by the parent). After `toggleEnvEnabled()` saves and sets `specOverride = result.spec`, the toggle's `aria-checked` attribute doesn't update because `envEnabled` still reads from the original prop, not from `specOverride`.

The save itself works (the API confirms the change), but the toggle visually stays in its previous position until the user does a full page reload.

### Suggested fix
Derive `envEnabled` from `specOverride` when available:
```js
let envEnabled = $derived(() => {
  const spec = specOverride ?? app.spec;
  const env = spec.environments?.find(e => e.name === currentEnv);
  return env?.enabled !== false;
});
```

Found during #228.

MC-Meesh:
Closing as **fixed** — verified against HEAD `5e3011e` in the mo-4jv quality audit. The `specOverride` variable this issue references no longer exists; the `acceptedSpec` refactor implements exactly the requested derivation:

- `ui/src/lib/components/drawer/SettingsTab.svelte:116-117` — `envEntry` derives from `(acceptedSpec ?? app.spec)`, so the toggle reflects the accepted server state immediately, no reload needed
- `toggleEnvEnabled` (`:135`) → `handleSpecUpdate` (`:99-103`) sets `acceptedSpec` and advances `acceptedGeneration`; the stale-prop guard at `:64-65` prevents regression
- Test coverage: `ui/src/lib/components/drawer/SettingsTab.test.ts:119-127,148-154`

PR #311, opened 2026-05-10
[codex] Fix release usability issues #279 #280 #281 #290 #295

MC-Meesh:
## Summary
This consolidates the fixes and regression coverage requested in hq-53t for issues #279, #280, #281, #290, and #295, plus the follow-up controller hardening needed after the first `#311` UI E2E CI failure.

## What changed
- Added a focused `UpdateApp` conflict-retry regression for `#279` proving the API survives a one-shot optimistic-lock conflict and persists the requested spec update.
- Fixed `#280` in the settings drawer by deriving the environment toggle state from the effective local spec (`specOverride ?? app.spec`) instead of stale props, and tightened the focused Playwright regression so the switch state must update immediately in the UI before any reload.
- Fixed `#281` by moving project environment creation to a retrying read-modify-write path against the live `Project`, with duplicate detection on each retry, and added focused envtest coverage that simulates a concurrent project update so the default `production` environment is preserved while `staging` is added.
- Hardened controller reconcile paths that were still dropping state under full-suite load by retrying `App` finalizer updates on conflict and treating concurrent project namespace creation as idempotent.
- Added controller tests covering finalizer update conflicts and concurrent namespace creation.
- Verified that the current `main` baseline already addresses `#290` via durable `BuildRun` state rather than in-memory-only build tracking, and already addresses `#295` via latched webhook registration/failure state keyed by stable inputs. This PR keeps those as verified regressions rather than duplicating existing tests.

## Root cause
- `#279`: the bug was previously caused by full-object update conflicts during concurrent controller writes; current server code already retries, but a focused regression test was missing.
- `#280`: the UI toggle rendered from `app.spec` while saves updated `specOverride`, so the persisted change could succeed while the switch stayed visually stale.
- `#281`: the create-environment path appended to a stale `Project.Spec.Environments` snapshot and called `client.Update`, which could drop concurrent controller/API changes.
- Follow-up `#311` CI failure: under full UI E2E load, controller reconcile conflicts around `App` finalizer updates and project namespace creation could abort downstream reconciliation work. That left later requests observing missing derived state, which matched the cross-signal integration failure where cloned env secrets stayed empty even though the cloned spec values were correct.
- `#290`: current `main` already persists build execution state through `BuildRun` objects.
- `#295`: current `main` already latches webhook registration success/failure in status conditions instead of retrying permanent failures forever.

## Validation
- `KUBEBUILDER_ASSETS=/Users/meesh/gt-codex/mortise/polecats/nitro/mortise/bin/k8s/1.35.0-darwin-arm64 go test ./internal/api -run '^(TestUpdateApp|TestUpdateAppRetriesConflict|TestCreateProjectEnvironment|TestCreateProjectEnvironmentPreservesConcurrentProjectUpdate|TestCreateProjectEnvironmentDuplicate|TestCreateProjectEnvironmentInvalidName|TestCreateProjectEnvironmentAsMemberForbidden)$'`
- `go test ./internal/controller -run '^(TestAddAppFinalizerWithRetry|TestRemoveAppFinalizerWithRetry|TestEnsureNamespaceHandlesAlreadyExistsFromConcurrentCreate)$'`
- `go test -tags=integration ./test/integration -run TestEnvironmentCloneCreatesEnvWithConfig -count=1`
- `npm run check` in `ui/`
- `CI=1 MORTISE_BASE_URL=http://localhost:8091 MORTISE_ADMIN_EMAIL=admin@local MORTISE_ADMIN_PASSWORD=admin123 npx playwright test`

## Polecat ownership
The code landed through real Gas Town polecat branches and was then integrated into crew3:
- `mo-wisp-eeg` / `nitro`: `#279` regression coverage
- `mo-wisp-0eh` / `rust`: `#280` UI fix and focused E2E tightening
- `mo-wisp-clk` / `chrome`: `#281` API fix and focused envtest coverage
- `mo-ab4` / `guzzle`: controller reconcile hardening after the draft PR CI failure

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -375,8 +375,9 @@ func main() {
 	}
 
 	if err := (&controller.ProjectReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:    mgr.GetClient(),
+		Scheme:    mgr.GetScheme(),
+		APIReader: mgr.GetAPIReader(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "Failed to create controller", "controller", "Project")
 		os.Exit(1)
--- a/internal/api/apps_test.go
+++ b/internal/api/apps_test.go
@@ -230,7 +230,7 @@ func TestUpdateAppRetriesConflict(t *testing.T) {
 	baseClient := fake.NewClientBuilder().WithScheme(scheme.Scheme).WithObjects(project, app).Build()
 	s := &Server{
 		client: &appUpdateConflictClient{Client: baseClient, fired: &fired},
-		authz:  allowAllPolicy{},
+		authz:  allowAllAppsPolicy{},
 	}
 
 	req := httptest.NewRequest(http.MethodPut, "/api/projects/default/apps/update-conflict", bytes.NewBufferString(`{"source":{"type":"image","image":"nginx:1.26.0"}}`))
@@ -260,9 +260,9 @@ func TestUpdateAppRetriesConflict(t *testing.T) {
 	}
 }
 
-type allowAllPolicy struct{}
+type allowAllAppsPolicy struct{}
 
-func (allowAllPolicy) Authorize(context.Context, auth.Principal, authz.Resource, authz.Action) (bool, error) {
+func (allowAllAppsPolicy) Authorize(context.Context, auth.Principal, authz.Resource, authz.Action) (bool, error) {
 	return true, nil
 }
 
--- a/internal/api/apps_test.go
+++ b/internal/api/apps_test.go
@@ -1,15 +1,24 @@
 package api
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
+	"github.com/go-chi/chi/v5"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/scheme"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	mortisev1alpha1 "github.com/mortise-org/mortise/api/v1alpha1"
+	"github.com/mortise-org/mortise/internal/auth"
+	"github.com/mortise-org/mortise/internal/authz"
 )
 
 func TestNormalizeRepoURL(t *testing.T) {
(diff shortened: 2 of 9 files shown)


JamesSprow (commented):
Requesting changes for one blocking correctness issue and one important coverage gap.

1. `internal/controller/project_controller.go:327-329`
   The new `AlreadyExists` branch in `ensureNamespace` recursively calls `ensureNamespace` immediately:

   ```go
   if errors.IsAlreadyExists(err) {
       return r.ensureNamespace(ctx, project, nsName, spec)
   }
   ```

   In the real controller this client reads through an informer cache, so the exact concurrent-create race is often:

   - cached `Get` still says `NotFound`
   - `Create` returns `AlreadyExists`
   - recursive retry hits the same stale cache again

   That means this can recurse/hot-loop until the cache catches up, and under bad timing it can become an unbounded recursion path inside one reconcile. The fake-client test added here does not model that because the created namespace is visible immediately.

   This needs to be changed to a non-recursive path that tolerates cache lag safely (for example: re-read via a fresh/API-backed client, or return and let the next reconcile observe the namespace).

2. `internal/controller/controller_conflict_test.go:87-128`, `internal/controller/app_controller.go:912-923`, `2432-2450`, `3369-3399`
   The PR adds new conflict-sensitive status update paths via `updateAppStatus`, but the new tests only cover finalizer updates and the idealized namespace-create race. There is still no direct proof that the new status-write retry behavior survives conflicts in:
   - `setFailedCondition`
   - webhook condition updates
   - external-app/domain-collision status updates

   Since the stated purpose here is controller hardening after CI flakes, those paths should be covered explicitly.

Architecture/docs check:
- I did not find any `CLAUDE.md` / `SPEC.md` / `ARCHITECTURE.md` drift in the design itself.
- I also did not find any public-behavior change that obviously requires doc/spec updates.

Validation notes:
- The new controller conflict tests pass on the PR head.
- I could not run the envtest-backed API tests locally because kubebuilder assets are not installed on this machine (`/usr/local/kubebuilder/bin/etcd` missing).

JamesSprow:
Review summary: `REQUEST_CHANGES`

The UI regression coverage and the API-side concurrent-update coverage are directionally good, and the PR stays within the project’s architecture rules. I do not see a docs/spec/architecture update requirement from the user-facing behavior here.

The blocking issue is in the new namespace race hardening:
- `internal/controller/project_controller.go:327-329` handles `AlreadyExists` by recursively calling `ensureNamespace` immediately.
- In the real reconciler that `Get` reads through the cache, so a concurrent create can produce `Get -> NotFound`, `Create -> AlreadyExists`, then another cached `Get -> NotFound` again.
- That makes this a potential unbounded recursion / hot-loop under cache lag, which is exactly the class of race this PR is trying to harden.

There is also still an important coverage gap:
- `internal/controller/controller_conflict_test.go` covers finalizer retries and an idealized namespace-create race with a fake client.
- It does not cover the new conflict-sensitive status-write paths in `app_controller.go` (`setFailedCondition`, webhook-condition updates, external-app/domain-collision status writes).
- It also does not model informer cache staleness for the `AlreadyExists` namespace case.

Test assessment:
- Focused controller tests on the PR head passed locally.
- I could not run the envtest-backed API tests here because kubebuilder assets are missing (`/usr/local/kubebuilder/bin/etcd` not present).

Net: the PR is close, but the namespace retry path needs to be made cache-safe before this is ready.

PR #321, opened 2026-05-11
[codex] fix #317 preview token cache isolation

### LINKED ISSUE #317: Preview backfill uses inconsistent git-token cache keys, causing wrong-token reuse across same-named apps
## Summary

Landed from PR #285 (`Fix git state sync for rebuilds and previews`) at merge commit `71f239a64c82e12d2fc542d410ce3318207b6b89`.

The new preview backfill path uses a different `gitTokenCache` key than the existing git build path.

## Findings

- High: `prepareGitSource()` stores the resolved token under `app.Namespace + "/" + app.Name`, but `syncOpenPreviews()` loads and stores under bare `app.Name`.
  - `internal/controller/app_controller.go:401`
  - `internal/controller/app_controller.go:419`
  - `internal/controller/app_controller.go:434`

## Impact

This has two effects:
- preview sync never reuses the token already resolved earlier in the same reconcile, so it does unnecessary extra token resolution work
- more importantly, apps with the same name in different projects can collide in the cache and use the wrong repo token for preview reconciliation

For example:
- `pj-a/web` can populate the bare-name cache entry `web`
- `pj-b/web` can later load that token and call forge APIs with the wrong credentials

That can break preview backfill with auth/repo-access errors or reconcile previews using another project or user’s repo access.

## Why This Violates Intended Behavior

PR #285 introduced preview backfill to make preview state converge from the repo side when webhook delivery was missed. Using a cross-project cache key in that path breaks isolation and makes preview reconciliation depend on unrelated apps that happen to share a name.

## Test Gap

The landed tests cover preview reconcile create/update/delete behavior, but there is no test that:
- exercises `prepareGitSource()` and `syncOpenPreviews()` together and asserts the same cache key is used
- creates two same-named apps in different namespaces/projects and proves their cached tokens cannot collide

## Recommended Direction

Use one namespace-qualified cache-key helper everywhere, for example:
- `app.Namespace + "/" + app.Name`

Then use that helper for both `Load` and `Store` in `prepareGitSource()` and `syncOpenPreviews()`.

JamesSprow:
## Summary

This fixes issue #317 by making git-token cache access use one namespace-qualified key everywhere preview sync touches it.

## Root cause

`prepareGitSource()` already cached the resolved token under `app.Namespace + "/" + app.Name`, but `syncOpenPreviews()` loaded and stored under bare `app.Name`.

That created two problems:
- preview sync could miss the token already resolved earlier in the same reconcile and do unnecessary fallback resolution work
- same-named apps in different control namespaces could share a bare cache key and reuse the wrong repo token during preview reconciliation

## What changed

- added a single `gitTokenCacheKey()` helper in `internal/controller/app_controller.go`
- switched `prepareGitSource()` to store with that helper
- switched `syncOpenPreviews()` to load and store with that same helper
- added regression tests covering:
  - prepare + preview sync sharing the same cache entry
  - same-named apps in different namespaces keeping distinct cached tokens

## Validation

Passed locally:
- `go test ./internal/controller -run 'TestSyncOpenPreviews(ReusesPrepareGitSourceTokenCacheKey|IsolatesGitTokenCacheKeysByNamespace)$'`
- `go test ./internal/previewsync`
- `go test ./internal/controller -run "$(rg -o '^func (Test[^ (]+)' internal/controller/*_test.go | cut -d' ' -f2 | grep -v '^TestControllers$' | paste -sd'|' -)"`

Local environment limitation:
- `go test ./internal/controller` still hits the envtest suite bootstrap and fails before specs run because this machine does not have the required envtest binaries (`../../bin/k8s` missing and `/usr/local/kubebuilder/bin/etcd` missing).

## Impact

Preview backfill now reuses the same in-reconcile git token cache entry as the build path and same-named apps in different projects cannot collide through bare app-name cache access.

### DIFF
--- a/internal/controller/app_controller.go
+++ b/internal/controller/app_controller.go
@@ -137,6 +137,9 @@ func (r *AppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		}
 		return ctrl.Result{}, err
 	}
+	cacheKey := gitTokenCacheKey(&app)
+	r.gitTokenCache.Delete(cacheKey)
+	defer r.gitTokenCache.Delete(cacheKey)
 
 	// Finalizer flow — owner references can't cross namespaces, so the only
 	// way to clean up per-env-ns resources on App delete is via a finalizer
--- a/internal/controller/app_controller.go
+++ b/internal/controller/app_controller.go
@@ -99,7 +99,8 @@ type AppReconciler struct {
 	Builds *BuildTrackerStore
 
 	// gitTokenCache holds the resolved git token for the current reconcile
-	// iteration so per-env builds don't re-resolve credentials. Keyed by app name.
+	// iteration so per-env builds don't re-resolve credentials. Keyed by the
+	// app's control-namespace/name pair.
 	gitTokenCache sync.Map
 
 	GitAPIFactory func(*mortisev1alpha1.GitProvider, string, string) (git.GitAPI, error)
@@ -433,9 +434,13 @@ func (r *AppReconciler) prepareGitSource(ctx context.Context, app *mortisev1alph
 	}
 
 	// Stash the token transiently for per-env builds during this reconcile.
-	r.gitTokenCache.Store(app.Namespace+"/"+app.Name, tokenResult.Token)
+	r.gitTokenCache.Store(gitTokenCacheKey(app), tokenResult.Token)
 
 	return ctrl.Result{}, true, nil
+}
+
+func gitTokenCacheKey(app *mortisev1alpha1.App) string {
+	return app.Namespace + "/" + app.Name
 }
 
 func (r *AppReconciler) syncOpenPreviews(ctx context.Context, app *mortisev1alpha1.App, project *mortisev1alpha1.Project) error {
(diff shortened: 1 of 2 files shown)


JamesSprow:
Review pass scope: namespace-qualified git token cache keys, same-name cross-project isolation, cache-key reuse consistency between build and preview sync, and controller coverage.

Findings:
- The original fix correctly unified preview-sync cache access onto the namespace-qualified key.
- I found one additional correctness issue: `gitTokenCache` lives on the reconciler, so a stale entry could survive across reconciles and let `syncOpenPreviews()` reuse an old token when `prepareGitSource()` was skipped because builds were already current.

Direct fix pushed in `8468b10`:
- clear the app cache entry at the start of `Reconcile()` and defer clearing it again on exit, so the cache actually stays reconcile-scoped
- add a controller regression test covering stale cache reuse across reconciles, in addition to the existing same-namespace key-reuse and cross-namespace isolation tests

Validation:
- `go test ./internal/controller -run 'Test(SyncOpenPreviews(ReusesPrepareGitSourceTokenCacheKey|IsolatesGitTokenCacheKeysByNamespace)|ReconcileClearsStaleGitTokenCacheBetweenReconciles)$'`
- `go test ./internal/previewsync`
- `make test`
- `make test-charts`

Escalations/open questions:
- none from this pass

CI:
- updated branch pushed; watching the new Actions run for the new head commit now

Issue #318, opened 2026-05-11
SSE token redeem skips auth revalidation, allowing post-revocation stream connects


JamesSprow:
## Summary

Landed from PR #272 (`SSE token exchange, JWT refresh, and client-side role trust`) at merge commit `b6bb84aad9a13f0b284a139e127b88d55cf21965`.

The new SSE token exchange stores a full `auth.Principal` at issue time and later redeems it without re-validating against the auth provider.

## Findings

- Medium: a user can fetch an `msse_` token, then be revoked or have their password generation bumped before the EventSource connection is opened, and the redeem still succeeds because the cached principal is accepted directly.
  - `internal/api/sse_tokens.go:50`
  - `internal/api/sse_tokens.go:67`
  - `internal/api/middleware.go:104`
- Low: the same merge left docs and generated API contract drift around the new SSE exchange and refresh behavior.
  - `docs/api-quickstart.md:155`
  - `docs/systems-overview.md:103`
  - `docs/swagger.yaml`
  - `internal/api/openapi.yaml`

## Impact

The `msse_` TTL is short, but the resulting SSE stream can be long-lived. A revoked user can still establish a fresh logs/events stream if they fetched the SSE token shortly before revocation.

Affected surfaces:
- project event stream
- app logs stream

## Why This Violates Intended Behavior

Before PR #272, the direct JWT-on-connect path re-ran auth-provider validation at stream-connect time. The new exchange path was meant to remove JWTs from query strings, not weaken revocation semantics.

## Repro Shape

1. Authenticate and call `POST /api/auth/sse-token`.
2. Revoke the user or change their password generation.
3. Connect to an SSE route with the issued `msse_` token.
4. Observe that the connection is still accepted.

## Test Gap

The landed tests cover revoked-before-issue and single-use behavior, but not revoke-or-password-change between issue and redeem.

## Recommended Direction

Do not cache a fully trusted principal for redeem.

Instead:
- store only the minimum claims needed to revalidate, such as `{email, passwordGen}`
- call back into the auth provider on redeem to obtain a fresh validated principal
- only then attach the principal to request context

PR #324, opened 2026-05-11
[codex] fix #319 refresh dead-code cleanup

### LINKED ISSUE #319: Remove shadowed /api/auth/refresh handler and realign refresh tests
## Summary

Landed from PR #269 (`JWT token refresh flow`) at merge commit `1253b58c90820aaf2ef1ca4e96a24d986a4ee446`.

The merge left two handlers registered for the same `POST /api/auth/refresh` path:
- unauthenticated top-level route in `internal/api/server.go`
- authenticated `/api`-group route (`RefreshToken`) in the same file

In practice, the new unauthenticated refresh handler is the one that matters, and the older `RefreshToken` implementation is effectively dead.

## Findings

- Low: duplicate route registrations leave `RefreshToken` effectively unreachable and make the older refresh tests misleading.
  - `internal/api/server.go:163`
  - `internal/api/server.go:175`
  - `internal/api/sse_tokens.go:140`
- The existing refresh tests still hit `/api/auth/refresh`, so they now validate the new top-level handler rather than the legacy implementation they appear to cover.
  - `internal/api/sse_tokens_test.go`

## Impact

Current runtime behavior is fine because the intended refresh handler is the new one, but the tree now contains:
- stale code
- misleading test coverage
- a higher risk of future edits landing in an unreachable path

## Why This Matters

This is not a user-facing outage today, but it is still actionable technical drift in an authentication surface. Leaving dead auth handlers and ambiguous test coverage around refresh logic is exactly how future auth regressions get hidden.

## Test Gap

The refresh tests no longer prove anything about the legacy `RefreshToken` path even though they look like they do.

## Recommended Direction

- remove the redundant `RefreshToken` route and handler, or give it a distinct path if it still has a real purpose
- move or rename the legacy refresh tests so they clearly cover the intended handler
- update auth-route docs so they describe the single supported refresh path

JamesSprow:
Fixes #319

## Summary
- remove the shadowed authenticated `/api/auth/refresh` route registration under the JWT-protected `/api` group
- delete the unreachable `RefreshToken` handler and the native-auth helper that existed only to support it
- drop the duplicate refresh tests from `internal/api/sse_tokens_test.go` so refresh coverage lives only with the active public refresh handler in `internal/api/auth_test.go`

## Root Cause
The JWT refresh flow introduced a new top-level `POST /api/auth/refresh` endpoint, but the older authenticated route and handler were left behind on the same path. Chi matched the intended top-level route first, so the legacy handler was dead while nearby tests still appeared to validate it.

## Scope
- refresh dead-code cleanup only
- no auth-flow redesign
- no SSE redeem revalidation changes pulled from #318 beyond re-reading that branch shape to avoid conflicting coverage

## Validation
- `go test ./internal/api -run 'Test(Refresh|IssueSSEToken|SSEToken)' -count=1` with `KUBEBUILDER_ASSETS=$(pwd)/bin/k8s/1.35.0-linux-amd64`
- `go test ./internal/api ./internal/auth -count=1` with `KUBEBUILDER_ASSETS=$(pwd)/bin/k8s/1.35.0-linux-amd64`


JamesSprow:
Reviewed the refresh-route cleanup and auth routing behavior.

Findings:
- No correctness regression in the cleanup itself.
- The branch was missing an explicit regression proving the surviving public refresh route still accepts recently expired JWTs.
- One middleware comment was stale after the route split.

Fixes made:
- Added `TestRefreshAllowsRecentlyExpiredToken` and tightened it to validate the newly issued JWT.
- Updated the middleware comment to match the actual protected/public route split.

Validation:
- `go test ./internal/api -run 'Test(Refresh|IssueSSEToken|SSEToken)' -count=1`
- `go test ./internal/auth -count=1`
- `make test`
- `make test-charts`

PR #348, opened 2026-05-11
[codex] Env redaction and secret guardrails

### LINKED ISSUE #313: App deploy tokens fail for project-declared environments unless the app already has an explicit env override
## Summary

Landed from PR #301 (`fix deploy token app and env validation`) at merge commit `0e5b77530a1b78f22c64900337a30b067934cbc3`.

`POST /api/projects/{project}/apps/{app}/tokens` now allows creating an app deploy token for any project-declared environment when the app participates in that env implicitly.

But `POST /api/projects/{project}/apps/{app}/deploy` still rejects the same env unless `app.spec.environments[]` already contains an explicit override entry for it.

This makes the newly-allowed token flow unusable for apps that rely on project-level environments without app-level override blocks.

## Findings

- High: `CreateToken` accepts project-declared envs without an app override, but `Deploy` still requires `app.spec.environments[i].name == req.environment` and returns `404 environment ... not found in app spec` otherwise.
  - `internal/api/tokens.go:92`
  - `internal/api/tokens.go:104`
  - `internal/api/deploy.go:105`
- The landed tests now assert token creation succeeds without an app override, but deploy coverage still only exercises envs that already have explicit overrides.
  - `internal/api/tokens_handler_test.go`
  - `internal/api/server_test.go`

## Why This Violates Contract

The App CRD contract says apps automatically participate in every project environment unless explicitly opted out; `spec.environments` carries overrides and opt-outs, not the complete env list.

The landed token behavior was updated to match that contract, but deploy handling still follows the older "explicit app override required" rule. That creates an API contract split inside the same app+environment deploy-token flow.

Relevant contract refs:
- `api/v1alpha1/app_types.go:411`
- `SPEC.md` deploy-token scope docs
- `docs/api-quickstart.md` token-then-deploy flow

## Repro

1. Create a project with `production` and `staging`.
2. Create an image-source app with no `spec.environments` entries.
3. Create a deploy token for `staging` via `POST /api/projects/{project}/apps/{app}/tokens`.
4. Use that token on `POST /api/projects/{project}/apps/{app}/deploy` with `{"environment":"staging","image":"nginx:1.27.2"}`.

## Actual

Token creation succeeds, but deploy returns:

`404 environment "staging" not found in app spec`

## Expected

If apps implicitly participate in project environments unless opted out, env-scoped deploy should also work for those implicit envs.

## Test Gap

There is no end-to-end API test that:
1. creates an app with no env overrides,
2. creates a token for a declared project env,
3. deploys with that token to the same env,
4. asserts success and materialization of the per-env image override.

## Recommended Direction

Keep token creation and deploy behavior aligned.

Most consistent direction:
- auto-create the per-env override entry when an env-scoped deploy targets an implicit project env.

Fallback alternative:
- reject token creation for implicit envs too.

### LINKED ISSUE #332: App env vars and project shared vars are returned in plaintext by read APIs
## Summary
`GET /api/projects/{project}/apps/{app}/env` and `GET /api/projects/{project}/shared-vars` return raw variable values, including obviously sensitive credentials.

## Repro
- Set app env vars such as `A5_0511_API_TOKEN=super-secret-token` and `A5_0511_DB_URL=postgres://...`.
- Set a shared var such as `A5_0511_SHARED_SECRET=shared-top-secret`.
- Read both endpoints.

## Actual
The API returns plaintext values.

## Expected
Sensitive values should be redacted, omitted, or otherwise protected by a stronger read model.

## Impact
This is direct secret exposure in normal read endpoints that back the variables UI.

### LINKED ISSUE #333: Reserved app runtime secret names can be user-created and then become undeletable
## Summary
The app secret create endpoint allows the reserved runtime secret name `{app}-env`, after which the secret disappears from list responses and cannot be deleted through the same API.

## Repro
1. Create app `a5-0511-hiddenapp`.
2. `POST /api/projects/{project}/apps/{app}/secrets?env={env}` with `name={app}-env`.
3. List app secrets.
4. Try to delete the same secret.

## Actual
- Create succeeds.
- List returns `[]`.
- Delete returns `403` because the secret is now treated as reserved runtime state.

## Expected
The API should reject reserved secret names up front.

JamesSprow:
## What changed
- Redacted env and shared-var values unless the caller is authorized for update access.
- Blocked creation of app runtime-reserved secret names.
- Added focused API coverage for env redaction and deploy-token behavior.

Closes #313.
Closes #332.
Closes #333.

## Why
- Prevents secret values from being exposed to read-only users.
- Protects Mortise-managed runtime secret names from user collision.
- Verifies deploy tokens can still target declared environments without explicit app overrides.

## Impact
- Developers and admins still see plaintext where authorized; viewers and restricted callers receive metadata-only responses.
- Secret creation now returns `409 Conflict` for reserved runtime secret names.
- Deploy-token coverage is stronger without changing unrelated API surfaces.

## Checks
- `KUBEBUILDER_ASSETS=/tmp/mortise-upstream-main-api-sweep-20260511/bin/k8s/1.35.0-linux-amd64 go test ./internal/api -run 'TestGetEnvAndSharedVarsRedactionForDeveloperAndViewer|TestCreateSecretRejectsReservedRuntimeSecretName|TestCreateTokenMissingAppReturns404|TestDeployTokenCanDeployToDeclaredEnvWithoutExplicitAppOverride'`

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -297,6 +297,8 @@ definitions:
     properties:
       name:
         type: string
+      redacted:
+        type: boolean
       source:
         type: string
       value:
--- a/internal/api/env.go
+++ b/internal/api/env.go
@@ -23,9 +23,10 @@ import (
 
 // envVarResponse is the JSON shape for a single env var.
 type envVarResponse struct {
-	Name   string `json:"name"`
-	Value  string `json:"value"`
-	Source string `json:"source,omitempty"`
+	Name     string `json:"name"`
+	Value    string `json:"value"`
+	Source   string `json:"source,omitempty"`
+	Redacted bool   `json:"redacted,omitempty"`
 }
 
 // patchEnvRequest is the JSON body for PATCH .../env.
@@ -70,10 +71,15 @@ func (s *Server) GetEnv(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, err)
 		return
 	}
+	plaintext, err := s.canAuthorize(r, authz.Resource{Kind: "app", Project: projectName, Environment: envName}, authz.ActionUpdate)
+	if err != nil {
+		writeJSON(w, http.StatusInternalServerError, errorResponse{"authorization check failed"})
+		return
+	}
 
 	resp := make([]envVarResponse, 0, len(envs))
 	for _, e := range envs {
-		resp = append(resp, envVarResponse{Name: e.Name, Value: e.Value, Source: e.Source})
+		resp = append(resp, envResponseVar(e, plaintext))
 	}
 	writeJSON(w, http.StatusOK, resp)
 }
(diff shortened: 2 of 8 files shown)


JamesSprow:
Second-pass review found one real out-of-scope follow-up that should stay separate from this PR’s scope: shared vars can still bypass restricted-environment secrecy for project developers. Filed as #351 for design and contract follow-up.

PR #349, opened 2026-05-11
[codex] Activity backfill and log stream errors

### LINKED ISSUE #334: Project creation events are lossy in activity history
## Summary
Successful project creates do not reliably leave a durable creation event in project activity/history.

## Repro
Create several projects in quick succession, wait for control namespaces to settle, then inspect `GET /api/projects/{project}/activity`.

## Actual
Some projects permanently lack a `project create` event even though creation succeeded.

## Expected
Every successful project create should emit exactly one durable creation event.

## Suspected cause
The activity write appears to race namespace creation.

### LINKED ISSUE #335: Log SSE returns a normal 200 stream when pod log open actually failed
## Summary
When a pod is stuck in `ErrImagePull` or similar startup failure, the live log endpoint still establishes a successful SSE stream and emits the failure as a normal `data:` frame.

## Actual
The response is `200 OK` with `Content-Type: text/event-stream`, and the body contains an ordinary stderr-style event like `error opening log stream: ... waiting to start: trying and failing to pull image`.

## Expected
The transport should distinguish stream-establishment failure from real application log lines, for example with a non-2xx response or an explicit typed SSE error event.

JamesSprow:
## What changed
- backfill the missing project create activity event during project reconcile and mark the project once recorded so it is not duplicated
- emit structured fatal SSE error events when a pod log stream fails to open
- add focused coverage for both behaviors, including a failing log-stream test harness that exercises the real stream-open error path

Closes #334.
Closes #335.

## Why
Some projects could miss their initial create activity entry, and log streaming startup failures were only exposed as plain text lines with no structured error metadata for the UI.

## Impact
Activity history now self-heals for older projects without duplicating create events, and the logs UX can distinguish fatal stream startup failures from normal log lines.

## Checks
- `go test ./internal/api -run '^TestLogsStreamOpenFailureUsesStructuredErrorPayload$'`
- `go test ./internal/controller -run '^TestControllers$' -ginkgo.focus='project create activity backfill'`

### DIFF
--- a/internal/api/logs.go
+++ b/internal/api/logs.go
@@ -30,6 +30,9 @@ type logLine struct {
 	Ts     string `json:"ts,omitempty"`
 	Line   string `json:"line"`
 	Stream string `json:"stream"`
+	Kind   string `json:"kind,omitempty"`
+	Code   string `json:"code,omitempty"`
+	Fatal  bool   `json:"fatal,omitempty"`
 }
 
 // parseLogLine splits a kubelet log line of the form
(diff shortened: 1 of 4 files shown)


JamesSprow:
Second-pass review found one real out-of-scope follow-up that should stay separate from this PR’s scope: the UI log viewer still ignores the new fatal structured SSE error fields and renders them like ordinary log lines. Filed as #352 for frontend follow-up.

PR #356, opened 2026-05-12
Normalize project not-found API responses

### LINKED ISSUE #341: Missing-project errors are inconsistent across routes
## Summary
Project CRUD and nested project routes return inconsistent not-found messages for the same missing project condition.

## Actual
- Top-level project CRUD can return raw Kubernetes strings such as `Project.mortise.mortise.dev "ghost" not found`.
- Nested routes already return cleaner API-level messages such as `project "ghost" not found`.

## Expected
Missing-project responses should be normalized across the API.

JamesSprow:
## Summary
- route top-level project CRUD through a shared project lookup helper
- reuse that helper for nested `resolveProject` and `getProject` paths so missing projects always return `project "name" not found`
- add route-level API coverage for top-level CRUD plus representative nested routes across both helper paths

## Testing
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" go test ./internal/api`

Closes #341.


JamesSprow (commented):
Reviewed against REVIEW.md checklist. No issues found — this is ready to merge.

**What I checked:**

- **Architecture (1):** No third-party SDK imports in controllers. No new interfaces or plugin patterns. N/A items confirmed.
- **REST API (8):** All four consolidated handlers (`GetProject`, `DeleteProject`, `UpdateProject`, `resolveProject`) retain their `authorize()` calls before `lookupProject`. Error responses remain structured JSON via `writeJSON`/`writeError`. No new endpoints.
- **Correctness (5):** `lookupProject` returns `*Project` (pointer). Callers in `UpdateProject` and `DeleteProject` correctly mutate through the pointer and pass it to `client.Update`/`client.Delete` without double-referencing. `resolveProject` correctly extracts namespace from the returned pointer. Nil/zero-value paths handled (returns `nil, false` on error).
- **Completeness:** Verified that `CheckProjectNameAvailable` (line 205) is intentionally excluded — it returns `available: true` on not-found, a different flow. Internal retry loops in `createProjectEnvironment`/`updateProjectEnvironment` correctly skip `lookupProject` since they re-fetch inside retry context.
- **Tests (11):** Table test covers all four consolidated paths (PATCH, DELETE, nested GET via `resolveProject`, nested GET via `getProject`). `TestGetProjectNotFound` extended to assert normalized error body. Sad path (raw k8s string leak) explicitly guarded. Tests are independent.
- **Code quality (12):** Dead `name := projectName` aliases removed. `writeProjectNotFound` is only called from `lookupProject` but acceptable as named intent. Every changed line traces to the normalization task.
- **Red flags:** None triggered. No hardcoded prefixes, no `time.Now()`, no `:latest` tags, no Playwright changes.

LGTM — would approve if not the PR author.

JamesSprow (commented):
Re-review including Section 14 (Documentation).

## Code: Clean

The refactoring is correct and well-tested:
- `lookupProject` properly consolidates the duplicated project-fetch logic into a single helper
- Error responses use structured `errorResponse` via `writeJSON` — no raw strings leaked
- Table-driven test covers all four entry points (GET, PATCH, DELETE, nested route) and asserts the raw k8s error string is NOT exposed
- Auth checks remain upstream of `lookupProject` in each handler — no regression
- Pointer semantics are correct throughout (`*Project` returned, callers pass directly)

## Section 14 (Documentation): Minor gap

**CHANGELOG.md** — This PR normalizes the error message format for project-not-found responses (previously some paths leaked raw k8s errors like `projects.mortise.mortise.dev "foo" not found`, now all paths return `project "foo" not found`). This is user-facing for CLI and API consumers.

Suggest adding an entry:

```markdown
## [Unreleased]

### Fixed

- **Project not-found errors**: All project API routes now return a
  consistent `project "name" not found` message instead of leaking raw
  Kubernetes error strings on some paths.
```

**docs/migration-v1.md** — Not required IMO. The HTTP status code (404) is unchanged, and the error message is now *more* user-friendly. This aligns with the existing migration table entry about internal errors being sanitized. Only consumers parsing exact error strings would notice, and those were never part of a documented contract.

This is a non-blocking nit — happy to approve once the CHANGELOG entry is added, or if you'd prefer to batch it with other pending changes.

PR #357, opened 2026-05-12
Handle fatal log stream SSE errors in both UI viewers

### LINKED ISSUE #352: LogViewer ignores fatal structured log stream errors from SSE
## Summary
The logs API now emits structured fatal SSE events when a pod log stream fails to open, but the current UI still treats those events as ordinary log lines.

## Current behavior
When `GET /api/projects/{project}/apps/{app}/logs` hits a stream-open failure, the API emits an event like:

```json
{"kind":"error","code":"stream_open_failed","fatal":true,"stream":"stderr","line":"error opening log stream: ..."}
```

The frontend log viewer currently ignores the `kind`, `code`, and `fatal` fields and renders the event like a normal log line.

## Why this is a problem
The backend contract now distinguishes fatal stream startup failures from normal log output, but the UI does not surface that distinction. Users therefore do not get the intended UX improvement from the API change.

## Repro outline
1. Open the log viewer for an app whose pod log stream fails to open.
2. Observe the SSE payload includes `kind=error`, `code=stream_open_failed`, and `fatal=true`.
3. Observe the UI still renders that payload as a regular line instead of a distinct fatal/error state.

## Notes
The API change is backward-compatible, so this is a follow-up UI integration issue rather than a blocker for the backend PR.

JamesSprow:
## Summary
- recognize fatal structured log-stream SSE payloads in both UI viewer paths
- render a distinct inline fatal banner while still showing the raw log line and pod context
- add focused component coverage for both viewers with a fake EventSource

## Testing
- npm run test:unit -- src/lib/components/LogViewer.test.ts src/lib/components/drawer/DeployLogsTab.test.ts
- npm run check

Closes #352


JamesSprow (commented):
## PR Review: Handle fatal log stream SSE errors in both UI viewers

Checked against REVIEW.md. The structure is clean -- shared `parseLogStreamEvent` + `isFatalLogStreamEvent` helpers, a reusable `LogStreamFatalBanner` component, and both viewer paths updated. Two issues to resolve:

---

### 1. Red Flag #9: Tests cover only the happy path

Both `LogViewer.test.ts` and `DeployLogsTab.test.ts` have a single test case each that verifies the fatal banner appears when `kind: 'error'` + `fatal: true`. There are no tests for:

- **Normal events do NOT show the banner** -- a non-fatal event should not render `role="alert"`. This is the obvious sad-path complement.
- **Partial fatal flags** -- `kind: 'error'` without `fatal: true`, or `fatal: true` without `kind: 'error'`. The `isFatalLogStreamEvent` function requires both, but no test proves the boundary.

The `isFatalLogStreamEvent` and `parseLogStreamEvent` functions are pure and would benefit from a small unit test file (`log-stream-events.test.ts`) covering these edge cases. That would also be the natural place to test non-JSON input, missing fields, etc. At minimum, add a sad-path assertion to each existing test (e.g., confirm `screen.queryAllByRole('alert')` has length 1, not 2, after emitting one fatal + one normal event).

Per REVIEW.md Red Flag #9, this blocks approval.

---

### 2. Behavioral change in `LogViewer.svelte` (Checklist 5: Semantic correctness / Behavior change without test change)

The old `LogViewer.onmessage` handler had `try { JSON.parse(...) } catch { /* Ignore malformed events */ }` and required `typeof parsed.pod === 'string'` -- non-JSON SSE messages were silently dropped. The new code delegates to `parseLogStreamEvent`, which returns `{ pod: '', ts: '', line: data }` for non-JSON input. This means **non-JSON SSE messages that were previously silently dropped by LogViewer will now appear as log lines**.

This may be intentional (aligning LogViewer with the DeployLogsTab behavior, which always had pass-through), but it's a semantic change with no corresponding test or mention in the PR description. Please confirm this is intended.

Also: the guard `if (!parsed || typeof parsed.pod !== 'string')` at line ~100 of `LogViewer.svelte` is now dead code -- `parseLogStreamEvent` always returns `pod` as a string when it returns non-null. Consider simplifying to just `if (!parsed)`.

JamesSprow (commented):
Re-review complete. All three items from the previous review are resolved. LGTM -- ready to merge.

**Previous review items:**

1. **Tests for helper functions** -- `log-stream-events.test.ts` has 9 unit tests covering both `isFatalLogStreamEvent` (true/false for all boundary combinations of kind and fatal) and `parseLogStreamEvent` (full JSON, minimal JSON, non-JSON, empty string, JSON without line field).

2. **Sad-path tests** -- Both `LogViewer.test.ts` and `DeployLogsTab.test.ts` now have a second test that sends a non-fatal error event (`kind: 'error'` without `fatal: true`) and asserts `queryByRole('alert')` is null.

3. **Dead code guard removed** -- The `typeof parsed.pod !== 'string'` check is gone from LogViewer.svelte. The entire manual JSON parsing block was replaced by `parseLogStreamEvent`, which always returns `pod` as a string.

**REVIEW.md checklist:**
- **UI:** No `page.route()` mocking. `getByRole('alert')` queries match by role only (no name parameter), so `exact: true` is N/A. Accessibility via `role="alert"` on the banner.
- **Tests:** Happy + sad paths covered. Unit tests beside the code. Independent (beforeEach/afterEach reset state).
- **Correctness:** `isFatalLogStreamEvent` correctly requires both `kind === 'error'` AND `fatal === true`. `parseLogStreamEvent` handles all edge cases.
- **Code quality:** No unrelated changes. Every line traces to the task.

**Minor observations (non-blocking, no action needed):**
- `parseLogStreamEvent` return type is `LogLineEvent | null` but it never returns `null` -- all paths return an object. Callers check for null defensively, which is harmless.
- Line 8 of `log-stream-events.ts` has a redundant `typeof record.line === 'string'` ternary (already guaranteed by the outer if-guard).

JamesSprow (commented):
Re-review complete. LGTM -- ready to merge.

**What I checked:**
- Shared `parseLogStreamEvent` / `isFatalLogStreamEvent` in `log-stream-events.ts` -- correct, well-typed, used by both LogViewer and DeployLogsTab
- `LogStreamFatalBanner.svelte` uses `role="alert"` (correct semantic)
- Tests cover both happy path (fatal banner shown) and sad path (banner absent for non-fatal errors)
- Dedicated unit tests for the parser cover: all fields, minimal fields, non-JSON fallback, empty string, JSON without `line`
- No `page.route()` mocking (unit tests mock at module boundary via `vi.mock`, which is fine)
- New `LogLineEvent` fields (`kind`, `code`, `fatal`) are all optional -- backwards-compatible
- No architecture violations, no third-party SDK imports in controllers

**Documentation note (non-blocking):** CHANGELOG.md does not yet exist on `main`, so this PR cannot add an entry. Once the changelog lands, this feature (fatal error banner in log viewers) should be recorded under the next release.

**Minor observation (non-blocking):** In `log-stream-events.ts`, line 8 (`const line = typeof record.line === 'string' ? record.line : data`) has an unreachable false branch since the outer `if` already guarantees `typeof obj.line === 'string'`. Harmless, but could be simplified to `const line = record.line as string`.

Issue #368, opened 2026-05-12
Clone environment PVC support


JamesSprow:
## Summary

When cloning environments (including for preview environments), PVCs are currently skipped. This means preview environments and manually cloned environments don't get storage volumes from the source environment.

## Context

The preview environment rework (#359-#366) makes previews real clone environments. The clone logic in `internal/api/project_envs.go` (`cloneEnvToApp`) copies env overrides (replicas, resources, env vars, bindings, etc.) but does not handle PVC cloning. Apps with `spec.storage` entries will get new empty PVCs in the cloned environment rather than copies of the source data.

## Desired behavior

- When cloning an environment, PVCs from the source env should be cloned into the target env
- This likely involves creating VolumeSnapshot from the source PVC and restoring into a new PVC in the target namespace
- Need to handle: snapshot class availability, cross-namespace restore, storage class compatibility
- Should work for both manual env clone and PR preview clone

## Considerations

- Not all storage classes support VolumeSnapshot — need graceful fallback (skip PVC clone, create empty)
- Snapshot restore can be slow for large volumes
- Some apps may not want data cloned (e.g., write-ahead logs) — may need per-storage opt-out
- Preview environments are ephemeral — snapshot lifecycle management

Related: #367 (manual preview triggers)

Issue #367, opened 2026-05-12
feat: manual preview environment creation and rebuild triggers


JamesSprow:
## Problem

Preview environments are currently only created via git webhooks (PR open/sync). There is no way to:

- Manually create a preview for a branch that doesn't have a PR open
- Manually trigger a rebuild of an existing preview without pushing a new commit

## Context

Filed during the preview environment rework planning. The rework makes previews real clone environments, but the creation trigger is still exclusively webhook-driven.

## Potential features

1. **Manual rebuild**: Button in the UI to re-trigger a build for the current SHA. Useful when builds fail due to transient issues (registry timeout, BuildKit OOM).

2. **Manual creation**: Create a preview from the UI by selecting a branch. Useful for testing branches before opening a PR, or for repos without webhook integration.

## Related

- Preview environment rework (previews as clone environments)
- BuildRun CRD (durable build tracking)

Issue #366, opened 2026-05-12
Make preview environments behave more like clone environments with editable overrides


JamesSprow:
## Problem

Preview environments currently sit in an awkward middle ground:

- they are not modeled like a true clone of an existing environment
- but users still expect them to behave like a real environment derived from a known source
- and when they need to adjust something for a specific preview, the product does not make that override model obvious

In practice, this creates repeated preview bugs around env vars, bindings, ports, and overall predictability.

## Expected

A preview environment should behave much more like a clone environment:

- start from a clear source environment
- inherit the source environment's effective config in a predictable way
- still allow manual preview-specific overrides, including env vars
- make it obvious which values are inherited versus manually changed on the preview

That would preserve the convenience of automatic PR previews while still letting users fix or tweak a specific preview when needed.

## Actual

Preview environments currently behave as a partial/special-case environment model. They do not feel like a real clone, and they are easy to end up in a state where:

- the source environment works
- the preview exists
- but key runtime config is missing or behaves differently

## Why this matters

A lot of preview-environment pain seems to come from the product model itself, not just isolated bugs. Treating previews as clone-like environments with manual overrides would make them much easier to understand and operate.

## Related

- #120
- #360
- #363
- #365

JamesSprow:
Closing manually: this was resolved by merged PR #369 (feat: rework preview environments as real clone environments).

Issue #365, opened 2026-05-12
Add explicit preview inheritance controls in the UI


JamesSprow:
## Problem

Preview (PR) environments support inheriting from a source environment, but there is still no clear UI for choosing and managing that inheritance.

In practice, this makes preview behavior hard to reason about:

- users cannot explicitly choose whether previews should inherit from `production`, `staging`, or another environment
- users cannot easily confirm what the current preview source environment is
- when preview behavior is wrong, there is no obvious UI path to inspect or correct the inheritance source

There used to be a broader issue mentioning missing inheritance UI, but this still needs a concrete product fix.

## Expected

Project settings should expose preview inheritance clearly in the UI:

- choose the source environment previews inherit from
- show the currently configured source environment
- make it obvious which settings apply to future preview environments

## Actual

The inheritance model exists in the backend/CRD, but the UI still leaves users without a clear way to set and understand preview inheritance.

## Why this matters

A lot of preview-environment breakage becomes much harder to diagnose when inheritance is implicit or hidden. If previews are meant to be derived from another environment, users need an explicit product control for that.

## Related

- #200
- #360
- #363

JamesSprow:
Closing manually: this was resolved by merged PR #369 (feat: rework preview environments as real clone environments).

PR #369, opened 2026-05-12
feat: rework preview environments as real clone environments

### LINKED ISSUE #359: PreviewEnvironment sync drops domain when project.preview.domain is unset, even if PlatformConfig.domain exists
## Problem

Preview environments are created without a domain unless the project explicitly sets `spec.preview.domain`.

In my cluster on `v1.0.1`, I had:

- `PlatformConfig.spec.domain = jazure.net`
- `Project.spec.preview.enabled = true`
- `Project.spec.preview.domain` unset
- one open PR on a project with two apps

Mortise created the `PreviewEnvironment` CRs, built images, and deployed the preview pods, but both preview CRs were created with an empty `spec.domain`.

Because `PreviewEnvironmentReconciler` only creates an Ingress when `pe.Spec.Domain != ""`, that meant:

- no preview Ingress
- no preview URL in status
- no usable preview hostname

## Expected

If `Project.spec.preview.domain` is empty, preview sync should still fall back to the platform domain and stamp something like:

- `{app}-{project}-pr-{number}.{PlatformConfig.spec.domain}`

That matches the fallback behavior implied elsewhere in the codebase and is what users expect when a platform domain is already configured.

## Actual

Preview sync creates the `PreviewEnvironment` with `spec.domain = ""`, so no Ingress or URL is reconciled.

## Source pointers

- `internal/previewsync/service.go`
  - `ResolveDomainTemplate()` returns `""` when the preview template is empty
  - `desiredPreviewEnvironment()` uses that result directly for `pe.Spec.Domain`
- `internal/controller/previewenvironment_controller.go`
  - only calls `reconcilePreviewIngress()` when `pe.Spec.Domain != ""`

There is also a separate fallback helper in `internal/controller/previewenvironment_controller.go`:

- `ResolvePreviewDomain(template, appName, projectName, prNumber, platformDomain)`

but that fallback does not appear to be used by the preview sync path that creates the CR.

## Repro

1. Set `PlatformConfig.spec.domain`.
2. Create a project with preview enabled but no explicit `spec.preview.domain`.
3. Open a PR for a git-backed app.
4. Inspect the generated `PreviewEnvironment`.

Observed result: `spec.domain` is empty and no preview Ingress is created.

## Impact

Preview environments can exist and even be `Ready`, but they are not routable unless the user happens to define `project.spec.preview.domain` explicitly.

JamesSprow:
## Summary

- **Preview environments are now real clone environments.** A PR open event creates a `pr-{number}` entry in `Project.Spec.Environments`, clones overrides from the source env, and lets the app controller's existing `resolveEnvs()` fan-out handle deployment.
- **PE CRD slimmed to thin coordinator** — only carries PR metadata and lifecycle phase.
- **One PE per PR (project-scoped)**, not one per app.
- **New `Environment.Branch` field** enables per-env git branch overrides.
- **Deleted `previewsync` package** — logic moved into PE controller.
- **UI updated**: source env selector in settings, PR envs in env switcher, previews page redirects.

**Net -2,800 lines.**

### Issues resolved

| Issue | Problem | Fix |
|-------|---------|-----|
| #359 | Domain fallback broken | Normal domain template with pr-42 as env name |
| #360 | PORT env var missing | App controller injects PORT; no separate preview path |
| #361 | Domains API rejects preview refs | resolveAppEnv accepts PR env names |
| #362 | Env switcher can't represent previews | Separate PR Environments section |
| #363 | One PE per app in multi-app | One PE per PR (project-scoped) |
| #365 | Inheritance implicit | Source env explicit in project settings |
| #366 | Previews don't feel like real envs | They ARE real environments |

### New issues filed
- #367 — Manual preview triggers
- #368 — Clone environment PVC support

## Test plan
- [x] go build, go vet — clean
- [x] Controller/API/Webhook tests — 618 passed
- [x] UI svelte-check — 0 errors
- [ ] CI: make test, make test-charts, integration

Closes #359, #360, #361, #362, #363, #365, #366


JamesSprow (commented):
## Review: LGTM (checked against REVIEW.md)

Reviewed the full diff against the REVIEW.md checklist. This is a clean, well-structured PR. Net -2,800 lines is impressive — the rework eliminates an entire package (`previewsync`) and ~900 lines from the PE controller by delegating build/deploy to the app controller's existing `resolveEnvs()` fan-out.

### Checked (all pass)

- **Architecture**: No third-party SDKs in `internal/controller/`. PE controller only imports Mortise packages, stdlib, and k8s libs. No Traefik-specific CRDs, no plugin architecture.
- **Namespace & Naming**: All namespace construction goes through `constants.ControlNamespace()`, `constants.EnvNamespace()`, `constants.PreviewNamespace()`. No hardcoded `pj-` prefixes.
- **CRDs**: New `Environment.Branch` field has `+optional` marker. `zz_generated.deepcopy.go` is regenerated correctly (`BotPR *bool` deep-copy, PE spec now value-copied since all pointer fields were removed). CRD YAML in `charts/mortise-core/crds/` is identical to `config/crd/bases/`.
- **Controllers**: Reconcile is idempotent (re-reconcile test verifies this). Uses `clock.Clock`, never `time.Now()`. `RetryOnConflict` wraps all read-modify-write operations on Project and App. Finalizer added before side effects, removed after cleanup.
- **Correctness**: `ensureAppEnvOverride` correctly handles the exists-but-branch-changed case. `cloneEnvironment` deep-copies slices and maps (tested with mutation check). Webhook `dispatchPREvent` correctly handles close-regardless-of-enabled-state. `resolveSourceEnv` prefers "staging", falls back to first non-production env.
- **Concurrency**: All Project and App mutations use `retry.RetryOnConflict`. Status updates use fresh-read-then-mutate pattern.
- **Helm**: No `Chart.yaml` version/appVersion edits. CRD YAML matches.
- **Tests**: Happy and sad paths covered extensively — PE controller tests cover: disabled project, missing project, restricted source env, omitted sourceEnv resolution, happy path with env clone + secret copy + branch override, multi-app projects, image-source apps (no branch set), SHA change idempotency, deletion cleanup, deletion when project already gone. Integration tests updated. Webhook tests updated for the new per-project PE model.
- **UI**: E2E test added for source env selector. No `page.route()` mocking. New button selectors use `exact: true` for "Save PR config".
- **Security**: No secrets logged, RBAC updated minimally (added update/patch for projects and apps on PE controller, which it now needs).
- **Red flags scan**: None found. No `time.Now()`, no `:latest` tags, no hardcoded prefixes, no IngressRoutes, no plugin architecture, no Chart.yaml edits.

### Minor observations (non-blocking)

1. **`getByRole` without `exact: true`** — `getByRole('button', { name: 'General' })` in the new E2E test lacks `exact: true`, but this follows the pre-existing pattern used 7 times in the same file. Not introduced by this PR.
2. **Shallow pointer copy in `cloneEnvironment`** — `Replicas (*int32)` and probe pointers (`*ProbeConfig`) are shallow-copied. Safe in practice because the clone is immediately persisted (round-trips through serialization break pointer sharing), but worth a follow-up to deep-copy for defense-in-depth.
3. **`ensureProjectEnv` skips `ValidateProjectEnvLengths`** — Validation happens when the project controller re-reconciles after the env entry is added. The failure path is indirect but functional. Consider adding an inline check for faster feedback.

Issue #455, opened 2026-08-09
Manual rebuild is not environment-selectable; rebuild markers consumed by first BuildRun


MC-Meesh:
## Summary

Split from #367 (and adjacent to #436): the shipped manual-rebuild endpoint is not environment-selectable, and rebuild request markers are consumed by whichever environment's BuildRun is created first. On any multi-env app, "Rebuild" applies to effectively one arbitrary environment; for preview environments it can't do the right thing at all.

## Evidence (HEAD `5e3011e`)

- `internal/api/rebuild.go:37-82` — `POST .../rebuild` has no `?environment=` parameter (contrast `Redeploy`, which accepts one at `:134`); it patches three **app-wide** annotations
- `internal/api/rebuild.go:84-119` — `resolveGitBranchHead` resolves `app.Spec.Source.Branch` (default `main`), never a preview env's PR branch
- `internal/controller/buildrun_support.go:356-362` — the markers are cleared immediately after the **first** successful BuildRun `Create`; later envs in the same reconcile read empty markers (`:239-243`) and get `Trigger=Auto` with no `noCache`
- `internal/controller/app_controller.go:746-750` — preview envs take their revision from the PreviewEnvironment, so the `mortise.dev/revision` annotation the endpoint writes is ignored for previews anyway

## Fix shape

1. Accept `?environment=` on Rebuild and resolve the branch/SHA per-env (PR branch for preview envs)
2. Scope marker consumption per-environment instead of clear-on-first (same root mechanism as #436's first-env-only bug — coordinate the fixes)

_Filed from the mo-4jv quality audit._

Issue #447, opened 2026-07-25
Controller OOM kills active builds and leaves them terminally failed


JamesSprow:
## Problem

The default controller limit in the chart was `512Mi`. Starting several preview builds concurrently caused the Mortise controller pod to be OOM-killed (`exitCode: 137`, `reason: OOMKilled`). Active BuildRuns subsequently became terminally failed with messages such as:

```text
build tracker lost after recovery attempt 2
```

The associated PreviewEnvironment remained failed until the completed failed BuildRun records were removed and the preview was recreated/retried manually.

## Reproduction

1. Use the chart default controller resources (limit 512Mi).
2. Create a preview containing several apps that build concurrently.
3. Observe the controller restart during builds.
4. BuildRuns become `Failed` with `build tracker lost after recovery attempt 2`; no automatic resumable retry occurs.

## Expected

Build state should be durable or safely resumable across controller restarts. At minimum, interrupted builds should transition to a retryable state and the controller should automatically recreate/retry them when BuildKit is healthy. The chart default should be evaluated for realistic concurrent-build memory use.

## Workaround

Raising the controller to `requests.memory=512Mi`, `limits.memory=2Gi` stopped the restarts in this cluster, but failed BuildRuns still required manual cleanup/retry.

MC-Meesh:
**Folding #290 (closed as duplicate) into this issue.** Combined acceptance criteria for the behavioral fix, with audit evidence at HEAD `5e3011e`:

1. **Interruption must be retryable, not terminal.** `internal/controller/buildrun_controller.go:174-180` fails the run with `BuildInterrupted` after `maxBuildRunRecoveryAttempts = 2`; `isTerminalBuildFailureCondition` (`app_controller.go:1233-1243`) latches it and `prepareGitSource` (`:623-625`) refuses all future builds until a manual rebuild — which is itself broken for multi-env apps (#436). The PE controller also hard-fails previews on it (`previewenvironment_controller.go:174-178,194-222`).
2. **Probe before relaunch (from #290).** On tracker loss, check whether the interrupted build's image already exists in the registry before re-running from scratch. `BuildRunReconciler.RegistryBackend` is declared (`buildrun_controller.go:57`) but only nil-checked (`:189`) — never queried.
[...]

PR #456, opened 2026-08-09
fix(build): BuildRun reliability — persist terminal status before tracker delete; rebuild markers reach all envs

### LINKED ISSUE #434: BuildRun tracker deleted before terminal status persists; restarts or wrongly fails completed builds
The durable BuildRun controller deletes its in-memory tracker before the terminal status is persisted, so a conflict on the status write can restart or wrongly fail a completed build.

In the terminal branch of `Reconcile`, `internal/controller/buildrun_controller.go:104` calls `r.Builds.delete(key)` **before** `persistBuildRunLog` (:105) and the terminal `Status().Update()`. Both can fail: the log ConfigMap upsert can conflict, and the status update runs against a `br` read from the informer cache that is routinely stale right after the previous reconcile wrote `phase=Running`. On any such error the reconcile returns with the tracker already gone and the API phase still `Running`, so the next reconcile enters `handleLostBuildRunTracker` (:142/:152). Because `StartedAt` is the original start time, the grace window has long elapsed, so it restarts a build that already succeeded (duplicate clone/build/push, result discarded); if `Attempt` is already 2 it marks a succeeded build `Failed/BuildInterrupted`, which then latches terminally on the App.

Related but distinct: on `IsNotFound` the reconcile returns immediately at `buildrun_controller.go:69-70` without cancelling the build goroutine or removing the `BuildTrackerStore` entry, so deleting a running BuildRun (or an App delete cascading its BuildRuns mid-build) leaks the tracker and lets the build run to completion (up to the build timeout) for an object that no longer exists.

**Fix:** delete the tracker only after the terminal status update succeeds; in the lost-tracker path re-read terminal state before restarting; on NotFound, look up the store entry, call `cancel()`, and delete it.

Related: #290 (in-memory tracker dropped on operator restart) is the restart-time variant of the same tracker/durable-status boundary.

---

### LINKED ISSUE #436: Manual rebuild only rebuilds the first environment of a multi-env app
A manual rebuild of a multi-environment app only rebuilds the first environment.

The API rebuild endpoint (`internal/api/rebuild.go`) sets app-wide `rebuild-requested-at`/`no-cache` annotations so the controller creates a fresh durable BuildRun even for the same SHA. But `ensureAppBuildRun` clears those markers (and mutates the in-memory `app` via `clearRebuildRequestMarkers`) as soon as the first environment BuildRun is created (`internal/controller/buildrun_support.go:357-363`). For environments 2..N in the same reconcile, `hasPendingRebuildRequest(app)` is now false, and the short-circuits in `reconcileEnvBuild` match the prior successful run (the selection hash deliberately ignores RequestID/NoCache/Trigger), so those environments never rebuild. This bites exactly the same-SHA and no-cache cases the annotation exists for.

**Fix:** clear the rebuild markers once after the environment loop completes, not inside `ensureAppBuildRun`.

---

MC-Meesh:
## Summary

Two BuildRun reliability fixes (Lane B of the 2026-08-09 audit, bead mo-lnl).

### #434 — BuildRun tracker race

**Half A — delete-before-persist.** The terminal branch of `BuildRunReconciler.Reconcile` deleted the in-memory tracker *before* the fallible log ConfigMap and status writes. Any failure there left the BuildRun at phase Running with no tracker, so the next reconcile treated a *finished* build as lost — restarting a duplicate full rebuild, or at attempt 2 latching a bogus terminal `BuildInterrupted` failure on an already-succeeded build. Now:

- The log ConfigMap and terminal status are persisted first (status write wrapped in `RetryOnConflict` with a re-read, matching `projectTerminalBuildRunStatus`); the tracker is dropped only after everything durable has landed.
- `upsertConfigMap` retries on conflicts and lost create races — it read-modify-writes the legacy app-scoped log ConfigMap shared by all envs of an app, so these races are real.
- `handleLostBuildRunTracker` re-reads the BuildRun and leaves terminal ones alone before deciding to restart.
- `mergedBuildRunLogLines` gained a full-suffix dedup so the now-reachable persist retry doesn't duplicate log lines.

**Half B — NotFound leak.** Reconciling a deleted BuildRun (e.g. App deletion cascading mid-build) returned without touching the tracker, leaking the build goroutine and its log ring (up to 2MB) until the 30min build timeout. The NotFound branch now cancels and removes the tracker via a new atomic `BuildTrackerStore.cancelAndDelete` helper.

### #436 — rebuild markers consumed by first env only

`ensureAppBuildRun` cleared the app-wide rebuild annotations mid env-loop, right after the first successful Create. The env loop shares one `*App`, so envs 2..N saw empty markers and short-circuited (`buildRunSelectionHash` deliberately ignores Trigger/RequestID/NoCache, so same-SHA rebuilds always matched the previous run) — a requested rebuild/no-cache rebuild never reached later environments. Now:

- The mid-loop clear is removed; markers survive the whole loop, so every env creates its rebuild run.
- `reconcileEnvBuild`'s previously dead `shouldClearNoCache` return is armed on the paths that consumed a pending rebuild request; the existing post-loop clear hook in the app controller (previously dead code) now fires once after all envs have consumed the markers.
- Deliberate, documented, and tested edge case: a mid-loop error return leaves the markers in place, so the rebuild stays pending across reconciles instead of being half-applied. The marker-consumption code stays app-scoped in a shape a per-env scope (#455) can extend.

### Test changes

- **`TestEnsureAppBuildRunCreatesAndClearsRebuildMarkers` pinned the wrong behavior and was deliberately changed** to `TestEnsureAppBuildRunCreatesRunAndPreservesRebuildMarkers`: it now asserts the markers survive `ensureAppBuildRun` (and that the run carries RequestID/NoCache).
- New multi-env regression test `TestReconcileEnvBuildRebuildRequestReachesEveryEnv` — placed in `buildrun_support_test.go` beside the other `reconcileEnvBuild` unit tests (the audit suggested `app_controller_test.go`, but that file is the ginkgo/envtest suite; this follows the repo's copy-the-nearest-test convention).
- New `TestBuildRunTerminalStatusPersistsBeforeTrackerDelete` (injects a status-write failure, asserts the tracker survives for retry and logs persist exactly once), `TestBuildRunReconcileCancelsTrackerWhenBuildRunDeleted`, and `TestHandleLostTrackerDoesNotRestartTerminalBuildRun`.
- All six new/updated tests were verified to fail against the pre-fix code.

`make test` fully green locally (unit + envtest + chart/UI checks); `go vet ./...` clean.

Fixes #434
Fixes #436

### DIFF
--- a/internal/controller/app_controller.go
+++ b/internal/controller/app_controller.go
@@ -765,6 +765,9 @@ func resolveEnvBuildIdentity(app *mortisev1alpha1.App, env mortisev1alpha1.Envir
 // the image to use for this env's deployment, or "" if a build is still in
 // flight (caller should skip deployment and requeue). statusDirty is true when
 // applyEnvBuildSuccess mutated app.Status and the caller must flush.
+// shouldClearNoCache is true when this env consumed a pending rebuild request;
+// the caller clears the rebuild markers once after the env loop, so a mid-loop
+// error return leaves them in place and the rebuild stays pending.
 func (r *AppReconciler) reconcileEnvBuild(ctx context.Context, app *mortisev1alpha1.App, projectionEnvOrder []string, envName, branch, revision string) (image string, requeue bool, statusDirty bool, shouldClearNoCache bool, err error) {
 	log := logf.FromContext(ctx)
 
@@ -830,6 +833,12 @@ func (r *AppReconciler) reconcileEnvBuild(ctx context.Context, app *mortisev1alp
 		}
 	}
 
+	// Reaching ensureAppBuildRun with markers pending means this env consumes
+	// the rebuild request: the ensured BuildRun carries its RequestID/NoCache.
+	// Report that so the caller clears the app-wide markers once after the
+	// env loop — clearing mid-loop would starve the remaining envs.
+	consumedRebuild := hasPendingRebuildRequest(app)
+
 	run, err := r.ensureAppBuildRun(ctx, app, envName, branch, revision, imageRef.Full, pullRef.Full)
 	if err != nil {
 		log.Error(err, "ensure buildrun failed", "env", envName)
@@ -840,15 +849,15 @@ func (r *AppReconciler) reconcileEnvBuild(ctx context.Context, app *mortisev1alp
 	switch run.Status.Phase {
 	case mortisev1alpha1.BuildRunPhaseSucceeded:
 		r.applyEnvBuildSuccess(ctx, app, projectionEnvOrder, envName, revision, run.Status.Image, run.Status.Digest, run.Status.DetectedPort)
-		return run.Status.Image, false, true, false, nil
+		return run.Status.Image, false, true, consumedRebuild, nil
 	case mortisev1alpha1.BuildRunPhaseFailed:
 		if err := r.setBuildFailureCondition(ctx, app, firstNonEmpty(run.Status.FailureReason, "BuildFailed"), run.Status.FailureMessage); err != nil {
 			return "", false, false, false, err
 		}
-		return "", false, true, false, nil
+		return "", false, true, consumedRebuild, nil
 	default:
 		r.markEnvBuildInProgress(app, envName, revision)
-		return "", true, true, false, nil
+		return "", true, true, consumedRebuild, nil
 	}
 }
 
--- a/internal/controller/build_tracker.go
+++ b/internal/controller/build_tracker.go
@@ -156,6 +156,19 @@ func (s *BuildTrackerStore) delete(key types.NamespacedName) {
 	s.trackers.Delete(key)
 }
 
+// cancelAndDelete atomically removes the tracker for the given key and cancels
+// its build goroutine, releasing the build context and log buffer. Used when
+// the BuildRun disappears mid-build (e.g. App deletion cascades to it).
+func (s *BuildTrackerStore) cancelAndDelete(key types.NamespacedName) {
+	v, ok := s.trackers.LoadAndDelete(key)
+	if !ok {
+		return
+	}
+	if t := v.(*buildTracker); t.cancel != nil {
+		t.cancel()
+	}
+}
+
 // GetBuildLogs returns the current build log lines for the given App, or nil
 // if no build is in progress. Exported for use by the API layer.
 func (s *BuildTrackerStore) GetBuildLogs(key types.NamespacedName) []string {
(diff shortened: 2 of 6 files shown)


MC-Meesh (commented):

Reviewed against the mo-4jv audit evidence for #434/#436 (lane B packet on mo-lnl). Both fixes match the evidence, and both tests that pinned the wrong behavior were rewritten deliberately with explanatory comments. Approving, with one race worth a follow-up and two observations.

**#434 half A** — terminal ordering is exactly right: log persist → status update in `RetryOnConflict` with a fresh `Get` → `projectTerminalBuildRunStatus` → tracker delete last. `upsertConfigMap` now retries both `Conflict` and `AlreadyExists` with the read inside the loop, which covers the shared legacy `buildlogs-<app>` ConfigMap race. The dedup guard in `mergedBuildRunLogLines` closes the double-append on persist retry, and the interceptor test asserts the lines land exactly once.

**#434 half B** — `cancelAndDelete` on the NotFound path releases the goroutine and the log ring; `LoadAndDelete` makes it atomic.

**#436** — moving marker consumption to a per-env `consumedRebuild` return and letting the existing post-loop consumer clear is the right shape. Verified the parts outside this diff: `clearNoCacheBuildAnnotation` clears all three markers via `clearRebuildRequestMarkers` (name is misleading, behavior correct), the loop ORs the flag rather than overwriting, every env consumes in the same pass because `requeue` uses `continue`, and a mid-loop error retry can't mint duplicate runs because `buildRunName` includes `RequestID` so `ensureAppBuildRun` re-finds the existing run.

**Follow-up worth filing (not blocking):** the re-read in `handleLostBuildRunTracker` uses the same cached client as the original read. Right after a terminal persist + tracker delete, informer lag can still present `Running`-with-no-tracker to a timer-triggered reconcile (e.g. a pending `buildRunPollInterval` requeue), and with `StartedAt` set the 5s grace is already elapsed — so the guard falls through to a restart. Consequence is a rare duplicate build (not a terminal failure, since the reorder makes the re-run persist correctly), but an uncached read (`APIReader`) here — or requeuing briefly instead of restarting when the re-read still shows Running — would close it completely.

Two observations, no action needed: (1) if `gcBuildRunHistory` fails after the tracker delete, the terminal path doesn't obviously re-run GC on the next reconcile — pre-existing shape, worst case one extra retained run; (2) the interruption-is-terminal latch (`BuildInterrupted` → `prepareGitSource` refusal) is intentionally untouched here — that's the follow-on behavioral bead (#447), which builds on this ordering fix as planned.

Issue #437, opened 2026-07-05
Workload update/delete and GC paths skip Mortise ownership checks


MC-Meesh:
Several controller paths adopt, overwrite, or delete same-named resources without checking Mortise ownership, violating the CLAUDE.md rule "Mortise owns only what it creates."

The newer ConfigMap and credentials-Secret paths correctly gate on `isMortiseManaged` (`app_controller.go` ~:2277, ~:2658), but the Deployment, CronJob, Service, Ingress, ServiceAccount, and PVC paths adopt any same-named object and overwrite it. The worst case is `deploymentSelectorMismatch` (`app_controller.go` ~:1476-1481), which **deletes** an existing Deployment with a mismatched selector without confirming Mortise created it. Separately, `gcAppAcrossEnvs` (`app_env_gc.go` ~:49-52) deletes cluster-wide by `app.kubernetes.io/name` + `mortise.dev/project` only: `app.kubernetes.io/name` is the most common label in the ecosystem, and the selector omits the `app.kubernetes.io/managed-by=mortise` label that every creation path stamps, with no namespace scoping.

**Fix:** add `managed-by=mortise` to the GC selectors, and an `isMortiseManaged` guard before the selector-mismatch delete and the workload update paths.

---
Found during a Fable 5 quality audit of the codebase (same audit that produced #433). Verified against the code at the current main.

MC-Meesh:
Completed on main: 43fe5b7 (ownership guards on the six workload adoption/update/delete paths, both after the read and inside the conflict-retry closure, via conflictIfUnmanaged → ResourceConflict condition) and 509d3bf (the seventh path, external-source Service type change). The GC half landed earlier in #457. Tests: resourceConflictError cases in app_controller_test.go. Issue was left open after the lane merged; closing with the evidence.

Issue #411, opened 2026-05-19
App containers missing PodSecurity restricted fields


JamesSprow:
## Summary

All app containers created by the operator trigger PodSecurity `restricted:latest` warnings because they lack the required security context fields:

```
would violate PodSecurity "restricted:latest":
  allowPrivilegeEscalation != false (container "portfolio" must set securityContext.allowPrivilegeEscalation=false),
  unrestricted capabilities (container "portfolio" must set securityContext.capabilities.drop=["ALL"]),
  runAsNonRoot != true (pod or container "portfolio" must set securityContext.runAsNonRoot=true),
  seccompProfile (pod or container "portfolio" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

This warning appears for every app on every reconcile cycle.

## Current impact

- INFO-level warning only — does not prevent deployment
- All apps deploy and run successfully despite the warnings
- Logged on every reconcile, adding noise to operator logs

## Discussion points

- Should the operator set a restricted security context by default on all app containers?
- `runAsNonRoot: true` could break images that require root (e.g., some database images, nginx default)
- Would this be opt-in per-app, or a platform-level default with per-app override?
- Is this better addressed as a PlatformConfig setting or an App spec field?

MC-Meesh:
**Direction ruling (project steering, 2026-08-09) — re-scoping this issue.** Decision: Mortise will not enforce a PodSecurity posture on user workloads. No default restricted context, no platform-wide toggle. Instead, the actual bug the audit surfaced here gets fixed:

**The operator actively clobbers user-applied security contexts.** `reconcileDeployment` and `reconcileCronJob` force `SecurityContext = nil` on every reconcile (`app_controller.go` ~:1599-1600, ~:1815-1816 at audit HEAD) — anyone who `kubectl patch`es a restricted context to satisfy PSA finds it silently reverted on the next reconcile. That is the opposite of coexistence.

**New scope for this issue — pass-through, not posture:**
1. Add `securityContext` (pod + container) to the App spec as plain pass-through fields — whatever the user declares lands in the Deployment/CronJob verbatim. No defaults, no validation beyond schema.
2. Stop force-nilling: the desired state is built from the App spec (nil if unset), and the existing comparison helpers (`normalizePodSecurityContext`/`securityContextsEqual`, currently dead code) do the rest unchanged.
3. Users who want PSA-restricted workloads set the fields; the docs hardening guide will carry a copy-paste restricted snippet.

[...]

PR #457, opened 2026-08-09
fix: controller quick fixes — domain-collision recovery, env seed race, GC scoping

### LINKED ISSUE #281: Project environment creation race: staging POST can clobber default production env

### LINKED ISSUE #438: App wedges in Failed after a domain collision with no retrigger path
When two apps request the same domain, the losing app is set to `Phase=Failed` and never recovers on its own.

On collision the App is set `Phase=Failed` and the reconcile returns `{}, nil` with no requeue (`internal/controller/app_controller.go` ~:310-324). When the conflicting app later releases the domain, the Ingress delete event maps to the other app (the handler uses the Ingress own labels), so this App is never re-enqueued; `shouldRefreshFailedAppStatus` only refreshes Failed status for build failures, so even the status pass is skipped. The app self-heals only at the next informer resync (hours) or a manual edit.

**Fix:** return a `RequeueAfter` on domain collision, or map Ingress deletions to apps carrying a `DomainCollision` condition so they re-reconcile when the domain frees up.

---

MC-Meesh:
Three controller quick fixes (Lane A of the 2026-08-09 audit, bead mo-paz).

## Domain-collision recovery (#438)

A domain collision previously returned `ctrl.Result{}, nil` mid-env-loop with `Phase: Failed`, which had four bad effects: no requeue (an app stuck in Failed forever, since nothing watches the foreign Ingress that owns the host), the `DomainCollision` condition was never cleared even after the domain freed up, build status accumulated for earlier envs was dropped before the batched flush, and every remaining env plus opted-out-env GC was skipped.

Now the collision is recorded and deferred past the loop: only the colliding env's Ingress is withheld, other envs reconcile normally, the App goes `Failed` with the aggregated condition **and requeues** (1m; build polling interval wins when a build is in flight). On the next clean pass the condition is removed and a status refresh lets the App leave `Failed`. The external-source path gets the same treatment.

## Env seed race (#281 residual)

`CreateProject` created a Project with no environments and left seeding `production` to the controller, whose guard was `len(envs) == 0`. A client adding an env between create and first reconcile (the UI POSTs `staging` right after create, error swallowed) suppressed the seed forever — the project never got `production`.

Both halves per the audit: the API now seeds `production` at create, and the controller guard becomes "no non-preview env present" (kept as a backstop for kubectl-created Projects; preview envs alone no longer suppress it).

## Finalizer GC scoping (#437, GC half)

`gcAppAcrossEnvs` listed eight resource kinds **cluster-wide** selecting only on `app.kubernetes.io/name` + `mortise.dev/project` — a user or foreign operator reusing those label keys would have their resources deleted on App delete. The GC now additionally requires `app.kubernetes.io/managed-by: mortise` (stamped by `appLabels` on everything Mortise creates) and only sweeps namespaces labelled as owned by the project (control + env + preview, the same label `deleteOwnedNamespaces` already scopes by). `gcOptedOutEnvs` gains the same managed-by guard. The managed-by pair moves to a shared `constants.ManagedByLabel`/`ManagedByValue`; `envstore` aliases it.

The adoption-guard half of #437 is deliberately not here — it collides with #411 and is tracked as phase 2.

## Tests

- envtest: collision defers, other envs keep reconciling, requeue interval returned, recovery clears the condition and phase (main + external-source paths)
- envtest: default-env seeding (empty, preview-only, non-preview-present) and API create seeding
- fake-client: GC deletes fully-labelled resources in project namespaces; skips unmanaged look-alikes and foreign namespaces (both GC entry points)
- test harness: suite namespaces now carry the project label the scoped GC selects on

`make test` fully green locally.

Fixes #438
Fixes #281
Part of #437 (GC scoping half only; adoption guards tracked on the issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   env vars (e.g. `DB_PASS` from `pg.password`). Includes new
   `GET /binding-keys` API endpoint and BindingsPicker UI rewrite.
 
+### Changed
+
+- **Finalizer GC now requires the `app.kubernetes.io/managed-by=mortise`
+  label and is scoped to project-labelled namespaces**: resources created
+  before the appLabels era (or by anything else that reuses Mortise's
+  name/project label keys without the managed-by label) are no longer
+  deleted by App-deletion GC. Operators relying on GC to clean up
+  pre-appLabels leftovers must remove those by hand once.
+
 ### Fixed
 
 - **`valueFrom.secretRef` resolution**: Field existed on the CRD but the
--- a/internal/api/projects.go
+++ b/internal/api/projects.go
@@ -16,6 +16,7 @@ import (
 	mortisev1alpha1 "github.com/mortise-org/mortise/api/v1alpha1"
 	"github.com/mortise-org/mortise/internal/authz"
 	"github.com/mortise-org/mortise/internal/constants"
+	"github.com/mortise-org/mortise/internal/controller"
 )
 
 // maxProjectNameLen caps the Project name so env namespaces (`pj-{name}-{env}`)
@@ -183,6 +184,14 @@ func (s *Server) CreateProject(w http.ResponseWriter, r *http.Request) {
 		},
 		Spec: mortisev1alpha1.ProjectSpec{
 			Description: req.Description,
+			// Seed the default env at create instead of leaving it to the
+			// controller: a client adding an env (e.g. the UI POSTing
+			// staging) between create and first reconcile would otherwise
+			// suppress the controller's seed and leave the project without
+			// a production env.
+			Environments: []mortisev1alpha1.ProjectEnvironment{
+				{Name: controller.DefaultProjectEnvironment},
+			},
 		},
 	}
 
--- a/internal/api/projects.go
+++ b/internal/api/projects.go
@@ -16,7 +16,6 @@ import (
 	mortisev1alpha1 "github.com/mortise-org/mortise/api/v1alpha1"
 	"github.com/mortise-org/mortise/internal/authz"
 	"github.com/mortise-org/mortise/internal/constants"
-	"github.com/mortise-org/mortise/internal/controller"
 )
 
 // maxProjectNameLen caps the Project name so env namespaces (`pj-{name}-{env}`)
@@ -190,7 +189,7 @@ func (s *Server) CreateProject(w http.ResponseWriter, r *http.Request) {
 			// suppress the controller's seed and leave the project without
 			// a production env.
 			Environments: []mortisev1alpha1.ProjectEnvironment{
-				{Name: controller.DefaultProjectEnvironment},
+				{Name: constants.DefaultProjectEnvironment},
 			},
 		},
 	}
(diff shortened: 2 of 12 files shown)


MC-Meesh (commented):

Reviewed against the mo-4jv audit evidence for #438/#281/#437-GC (lane A packet on mo-paz). All three fixes match the evidence, scope discipline is right (#437's adoption guards correctly left to the phase-2 bead), and the deviations from the packet's suggestions are improvements. Verified off-diff assumptions where the fix depends on them.

**#438** — deferring the collision to after the env loop fixes all four defects at once: both sites (main loop and `reconcileExternalSource`) now requeue (`domainCollisionRequeueInterval`, with `buildPollInterval` taking precedence when a build is in flight); the condition is cleared on a clean pass in both paths; and the mid-loop return is gone, so build status flush, marker clear, and opted-out-env GC all run before the deferred failure — I verified the insertion point sits after all three. Withholding only the colliding env's Ingress while letting its workload and every other env reconcile is better than the packet's continue-with-deferred-failure sketch. Using `domainCollisionCleared` to force the status refresh instead of extending `shouldRefreshFailedAppStatus` is an equivalent, cleaner shape — accepted deviation. Poll-based recovery (1min) rather than an Ingress-delete watch mapping is the packet's recommended simple option; the constant's comment explains why polling is required.

**#281** — both halves as specified: API-side seed at create closes the window entirely, and the controller guard is now "no non-preview env present" with the preview-only case tested (append, not replace — the preview env survives). The kubectl-created-Project seed path is preserved and documented. One nit: `internal/api` now imports `internal/controller` just for `DefaultProjectEnvironment`. No import cycle (controller doesn't import api), but this repo keeps shared naming in `internal/constants` — consider moving the constant there before merge or in a follow-up, so the api→controller edge doesn't become a precedent.

**#437-GC** — the selector now requires `managed-by=mortise` and iterates only namespaces labelled `mortise.dev/project=<proj>`. I verified the load-bearing assumption: `namespaceLabels` in the Project controller stamps `constants.ProjectLabel` on every namespace it creates (control, env, and preview via the PE controller), and `gcStaleEnvNamespaces`/`deleteOwnedNamespaces` already select on exactly this label — so the scoping is consistent with existing machinery, not a new convention. Hoisting `ManagedByLabel`/`ManagedByValue` to `internal/constants` with envstore aliasing is exactly what the packet asked for. The two scoping tests (label-based and namespace-based survival) cover both escape hatches.

One migration note worth a line in the PR description: resources created by Mortise versions that predate `appLabels` stamping `managed-by` will now survive finalizer GC (leak on App delete) instead of being cleaned up. Current `appLabels` has stamped it for a long time, so the exposure is old installs only — but it's a behavior change on upgrade and belongs in the changelog entry.

Merge-train note: this PR and #456 both touch `app_controller.go` in adjacent regions (this one inserts after the post-loop consumer that #456's marker fix depends on; #456 edits `reconcileEnvBuild` returns). Ranges are disjoint — rebase should be clean in either order.

Issue #446, opened 2026-07-25
helm upgrade can remove registry-proxy despite registry.enabled=true


JamesSprow:
## Problem

Upgrading an existing Mortise release with the public chart removed the `registry-proxy` DaemonSet even though release values still contained:

```yaml
registry:
  enabled: true
  proxy:
    hostPort: 30500
```

After the upgrade, `helm get manifest mortise -n mortise-system` returned no manifest content, while the prior release manifest included `mortise/templates/registry-proxy.yaml`. The DaemonSet disappeared, so every workload using `localhost:30500/...` failed with:

```text
dial tcp [::1]:30500: connect: connection refused
```

## Reproduction

1. Install a release with registry proxy enabled.
2. Run:

```sh
helm upgrade mortise mortise/mortise -n mortise-system \
  --version 0.1.0 --reuse-values
```

3. Verify `registry.enabled` remains true with `helm get values -a`.
4. Observe that `registry-proxy` is no longer present and `helm get manifest` is empty.

## Expected

The published chart should render the same enabled components on an upgrade and preserve the registry proxy. Please add a chart upgrade/render regression test.

MC-Meesh:
**Audit findings (mo-4jv, HEAD `5e3011e`)** — the source templates cannot produce the reported behavior, which points the investigation at the publish layer:

- `charts/mortise/templates/registry-proxy.yaml:1` guards the DaemonSet with a plain `{{- if .Values.registry.enabled }}` — no hook, no lookup, no dependency condition. Statically there is no install-vs-upgrade difference in source.
- The likely real mechanism: **published-artifact drift**. `.github/workflows/release.yml:146-158` stamps `Chart.yaml` from the git tag at release time, so the in-repo `0.1.0` is a placeholder — and the `0.1.0` on gh-pages is an ancient artifact (latest release is 1.0.4). The reporter pinned `--version 0.1.0`; upgrading onto that legitimately drops templates added later. This is the **same mechanism** as the 2026-07-24 regression comment on #379 (published 0.1.0 again missing BuildRun RBAC), so one fix covers both: post-publish verification of the packaged umbrella chart in `release.yml`, plus a chart-version policy so `0.1.0` isn't a reusable placeholder.
- The regression test this issue asks for still doesn't exist: `test/chart/run.sh` has upgrade cases but only toggles `buildkit.enabled`, never uses `--reuse-values`, and never asserts the `registry-proxy` DaemonSet exists at any point.

[...]

Issue #454, opened 2026-08-09
install.sh still sets renamed metricsServer.enabled key — script silently no longer disables metrics-server


MC-Meesh:
## Summary

The official install script still passes `--set metricsServer.enabled=false`, but the values key was renamed to `metrics-server` in 1.0.4. Helm silently ignores unknown keys, so **the install script no longer disables metrics-server** — users who pick "my cluster already has metrics-server" get a second, conflicting metrics-server installed anyway.

## Evidence (HEAD `5e3011e`)

- `scripts/install.sh:253` — `--set metricsServer.enabled=false`
- `docs/install.md:394` — documents the old `metricsServer.enabled` key
- `CHANGELOG.md:28-33` — records the `metricsServer` → `metrics-server` rename in 1.0.4
- `charts/mortise/values.yaml` — has no `metricsServer` key (only `metrics-server`)

## Fix

- Update `scripts/install.sh:253` and `docs/install.md:394` to `--set "metrics-server.enabled=false"` (note: hyphenated keys need quoting in `--set`)
- Grep for any other stale `metricsServer` references
- Consider a `values.schema.json` (or a CI grep) so renamed keys fail loudly instead of being silently ignored — this is the second silent-rename casualty class after the chart-drift issues (#446/#379)

_Found during the mo-4jv quality audit while verifying #446._

PR #458, opened 2026-08-09
ci: pin lint tools, unify helm --wait budgets, fix k3d import on containerd store

### LINKED ISSUE #443: CI hygiene: run/pin linters, unify helm --wait budgets, k3d checksum + install drift
A handful of CI workflow issues make the pipeline less reliable and less reproducible. (#433 already adds the failure diagnostics and artifact upload; these are the remainder.)

- **golangci-lint is never run in CI.** The `lint` job runs `go vet` + `dominikh/staticcheck-action` only, while the repo maintains `.golangci.yml` and `.custom-gcl.yml` plus a golangci-lint Makefile target. Either run it in CI or delete the config.
- **Unpinned tool version:** the staticcheck action uses `version: latest`, so a new staticcheck release can break CI with no code change. Pin it.
- **helm `--wait` budgets are tight:** integration uses 180s, dev-up 300s (a failing run expired at exactly 301s), while pulling traefik/cert-manager/metrics-server images cold inside k3d on hosted runners. Raise to a uniform ~600s and pre-import chart-dependency images like the test images already are.
- **k3d checksum drift:** the integration job verifies the k3d binary sha256 (`ci.yml` ~:71) but the ui-e2e job downloads the same binary with no checksum (~:142-143). Copy the sha256sum line.
- **Install drift:** `make test-e2e` uses `npm install` while the ui job uses `npm ci`; Playwright browsers are re-downloaded every run (no `~/.cache/ms-playwright` cache).

**Fix:** run golangci-lint (pinned) or remove the config; pin staticcheck; unify helm timeouts and pre-import dep images; add the k3d checksum to the e2e job; switch `test-e2e` to `npm ci` and cache the Playwright browser.

---

MC-Meesh:
## Summary

CI-hygiene batch from the 2026-08-09 audit (#443 items 2–5) plus the k3d containerd-store import fix (tracker bead mo-29g).

- **Pin lint tools + drop staticcheck-action**: the lint job now runs `go install honnef.co/go/tools/cmd/staticcheck@v0.7.0 && staticcheck ./...` directly — pinning what `latest` resolved to in the last green run AND removing the action whose internal cache collides with setup-go's (the deterministic 43k-tar-error failure that hit #456 twice). The `sigs.k8s.io/logtools` golangci plugin is pinned `latest` → `v0.10.1`. A new upstream release can no longer break CI with no code change.
- **Unify helm `--wait` budgets**: one 600s constant — `HELM_WAIT_TIMEOUT` in the Makefile (dev + integration installs), mirrored in `test/chart/run.sh` (4 sites) and `scripts/install.sh`. Replaces the scattered 180s/300s values; a real dev-up run has expired at exactly 301s while pulling dep images cold.
- **Pre-import chart-dependency images**: traefik `v3.3.3` and metrics-server `v0.7.2` (versions from the packaged subcharts) join the fixture images in the k3d import list, keeping cold-registry pulls out of the `--wait` window.
- **k3d checksum drift**: the ui-e2e job now verifies the same sha256 the integration job already checks.
- **Install drift**: `make test-e2e` uses `npm ci` (was `npm install`); Playwright browsers cached keyed on `ui/package-lock.json`.
- **containerd-store import fix (mo-29g)**: with Docker's containerd image store (default on fresh Docker 28+), `docker save` of pulled multi-arch images emits tarballs referencing blobs they don't contain; the node-side `ctr import` fails — and k3d logs the ERRO but still exits 0, so the failure only surfaces later as `helm --wait` timing out on `ErrImageNeverPull` pods. The Makefile now detects the store via `docker info` and imports per-image with `docker save --platform linux/amd64 | ctr -n k8s.io images import -`; the legacy store (GitHub runners) keeps the plain `k3d image import` path.

**Deliberately excluded**: #443 item 1 (golangci-lint is never run in CI) — unknown first-run backlog; tracked separately as bead mo-6eh.

## Verification

- `make test` green (unit + envtest + svelte-check)
- `make test-charts` green
- `make test-integration` green end-to-end on a containerd-store Docker 29.1.3 host — 53/53, with the per-image import path exercised (previously this host could not run the suite at all)

Fixes #443

## Coordination notes

- Owns `ci.yml`, `Makefile`, `test/chart/run.sh`, `.custom-gcl.yml` per lane F; #433's CI diagnostics blocks untouched.
- `run.sh` timeout is a single `HELM_WAIT_TIMEOUT` env-overridable constant for the future #446 publish-drift lane to reuse.

### DIFF
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -8,4 +8,4 @@ plugins:
   # logcheck validates structured logging calls and parameters (e.g., balanced key-value pairs)
   - module: "sigs.k8s.io/logtools"
     import: "sigs.k8s.io/logtools/logcheck/gclplugin"
-    version: latest
+    version: v0.10.1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,9 +101,13 @@ jobs:
         run: go vet ./...
 
       - name: staticcheck
-        uses: dominikh/staticcheck-action@v1
-        with:
-          version: "v0.7.0"
+        # Direct pinned install instead of dominikh/staticcheck-action: the
+        # action's internal cache collides with setup-go's cache:true
+        # (deterministic 'File exists' tar errors on restore), and setup-go's
+        # module cache already covers the install.
+        run: |
+          go install honnef.co/go/tools/cmd/staticcheck@v0.7.0
+          staticcheck ./...
 
   ui:
     name: UI typecheck + build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
       - name: staticcheck
         uses: dominikh/staticcheck-action@v1
         with:
-          version: latest
+          version: "v0.7.0"
 
   ui:
     name: UI typecheck + build
@@ -152,7 +152,14 @@ jobs:
       - name: Install k3d
         run: |
           curl -fsSL -o /usr/local/bin/k3d "https://github.com/k3d-io/k3d/releases/download/v5.8.3/k3d-linux-amd64"
+          echo "dbaa79a76ace7f4ca230a1ff41dc7d8a5036a8ad0309e9c54f9bf3836dbe853e  /usr/local/bin/k3d" | sha256sum --check
           chmod +x /usr/local/bin/k3d
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ hashFiles('ui/package-lock.json') }}
 
       - name: Start dev cluster
         run: make dev-up
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,11 @@ DEV_IMG ?= mortise:dev
 DEV_OBSERVER_IMG ?= mortise-observer:dev
 GITHUB_CLIENT_ID ?= Ov23lizLTd25E32VrWwl
 K3D_RUNTIME_ULIMIT ?= nofile=1048576:1048576
-DEV_HELM_TIMEOUT ?= 300s
+# Single budget for every helm --wait in this repo (#443): hosted CI runners
+# pull traefik/cert-manager/metrics-server images cold, and 180-300s budgets
+# have expired in real runs.
+HELM_WAIT_TIMEOUT ?= 600s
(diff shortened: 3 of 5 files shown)


MC-Meesh (commented):

Reviewed against the mo-4jv audit evidence for #443 items 2-5 (lane F packet on mo-qsf), including the follow-up commit replacing the staticcheck action. All four items land as specified, item 1 (golangci-lint enablement) is correctly excluded per the packet, and the mo-29g containerd-store fix is a real root-cause fix with a well-written justification comment.

**Item 2 (pins):** logtools plugin pinned to v0.10.1, and staticcheck now a direct `go install …@v0.7.0` — the rationale comment for dropping the action (cache collision with `setup-go`'s `cache: true`) is exactly the kind of why-comment this repo wants. **Item 3 (budgets):** single `HELM_WAIT_TIMEOUT ?= 600s` constant, env-overridable, threaded through all six former sites — Makefile, both `run.sh` upgrade cases, the standalone install, and `install.sh`. Verified no stray `--timeout 180s`/`300s` remains in the diff scope. **Item 4:** the ui-e2e k3d download now carries the identical sha256 as the integration job's (verified byte-for-byte against main). **Item 5:** `npm ci` in `test-e2e` and a Playwright browser cache keyed on `ui/package-lock.json`.

**mo-29g:** the detection (`DriverStatus` contains `io.containerd.snapshotter`) plus per-image `docker save --platform | ctr -n k8s.io images import` on the containerd store — while keeping plain `k3d image import` on the legacy store — matches the failure mode described in the bead, and the comment explains the silent-success trap (`k3d` exits 0, failure surfaces later as `ErrImageNeverPull`).

Three non-blocking notes:
1. `INT_DEP_IMAGES` pins traefik v3.3.3 / metrics-server v0.7.2 copied from the packaged subcharts. If a chart bump changes those versions, the pre-import silently stops helping (cold pull returns, now inside a 600s budget — degradation, not breakage). A follow-up drift check comparing `INT_DEP_IMAGES` against the packaged subchart values would keep the optimization honest; fits naturally next to `verify-chart-dependency-drift`.
2. The ctr import path targets `k3d-$(INT_CLUSTER)-server-0` only. Correct for the single-node integration cluster; if the cluster ever grows agents, images land on one node and the failure will look exactly like mo-29g again. Worth one sentence in the comment.
3. Same-file adjacency: #454 (install.sh still passes the renamed `metricsServer.enabled` key, so it silently stopped disabling metrics-server) touches the file this PR already edits. Fine to keep separate for lane hygiene — just don't let it fall through the cracks.

Issue #435, opened 2026-07-05
Multi-repo preview environments collide on shared pr-{N} env and namespace


MC-Meesh:
In multi-repo projects, preview environments from different repos collide on a shared `pr-{N}` environment and namespace.

Preview environment object names are repo-disambiguated via `constants.PreviewEnvironmentName(repo, number, multiRepo)`, but the environment name is always `fmt.Sprintf("pr-%d", pe.Spec.PullRequest.Number)` (`internal/controller/previewenvironment_controller.go:93`, and the comparison at :909). PR numbers overlap across repos constantly, so two repos in one project with same-numbered open PRs produce two live PEs sharing environment `pr-N` and namespace `pj-{project}-pr-N`.

Damage: (a) `ensureAppEnvOverride` runs per-PE over every app in the control namespace; for an app in repo1, PE-A sets the branch override while PE-B (which does not target that repo) clears it, and each App update re-enqueues all PEs, producing an unbounded update loop that alternates rebuilds between the PR branch and the default branch; (b) `cleanupPreview` on the first PE closing removes the shared environment, strips overrides, and deletes the shared namespace while the other PE is still open.

**Fix:** include the repo slug in the environment name when `multiRepo` (mirroring the PE naming), or refcount environment/namespace cleanup so it only runs when the last PE referencing it closes.

---
Found during a Fable 5 quality audit of the codebase (same audit that produced #433). Verified against the code at the current main.

PR #459, opened 2026-08-09
fix(preview): refcount shared preview env cleanup and stop branch flip-flop loop

MC-Meesh:
Two same-numbered PRs in different repos of one project resolve to the same `pr-{N}` env name and namespace. This is phase L1 of #435 — stop the damage; the full env-name qualification is tracked on the issue.

- `ensureAppEnvOverride` no longer clears an override branch it does not own: a PE that doesn't target the app's repo skips the write instead of resetting the branch, breaking the set/clear flip-flop loop driven by the App watch.
- `cleanupPreview` skips `removeProjectEnv`/`removeAppEnvOverride`/`deletePreviewNamespace` while another live PE in the namespace resolves to the same env name, so closing one PR no longer tears down the other's running preview.
- The test that asserted the clearing behavior is deliberately rewritten to assert preservation; adds a flip-flop regression test and a refcounted-cleanup envtest.

Part of #435 (does not close it).

### DIFF
--- a/internal/controller/previewenvironment_controller.go
+++ b/internal/controller/previewenvironment_controller.go
@@ -493,6 +493,13 @@ func (r *PreviewEnvironmentReconciler) cleanupPreview(ctx context.Context, pe *m
 		return err
 	}
 	if inUse {
+		// Deliberate last-close-wins tradeoff: skipping ALL cleanup means the
+		// closed PR's own app keeps its pr-{N} override, so its workload keeps
+		// running from the closed branch until the last same-numbered PE
+		// closes. Removing just that app's override instead would rebuild the
+		// app from its default branch INTO the shared preview env — a worse
+		// lie than a briefly-stale zombie. Revisit if the L2 env-rename gives
+		// same-numbered PRs distinct env names.
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


MC-Meesh (commented):

Reviewed against the mo-4jv audit evidence for #435-L1 (lane C packet on mo-l1g). Both stop-the-bleeding fixes land exactly where the evidence pointed, and the test that asserted the old clearing behavior was rewritten deliberately with a comment explaining why. Approving, with one behavioral note to make an explicit decision on and one L2 breadcrumb.

**Flip-flop** — the ownership rule (`!previewTargetsAppRepo` → return before any write) kills the closed loop at its source. Verified `previewTargetsAppRepo` subsumes the dropped `SourceTypeGit` check (`preview_repo_match.go` returns false for non-git/empty-repo apps), so image-source apps are safely skipped too. The convergence test asserting unchanged `resourceVersion`s across repeated reconciles of both PEs is exactly the right regression shape — it catches any reintroduced write, not just branch changes.

**Refcounted cleanup** — `envInUseByAnotherPreview` (skip self, skip deleting) gating the whole teardown matches the packet, and the envtest covers both halves: shared resources survive the first close, full teardown on the last.

**Decision to make explicit (fine to defer, but say so in the code or bead):** with skip-everything cleanup, the first-closed PE's own app keeps its `pr-{N}` override — repo-A's preview workload keeps running from the closed PR's branch until the last same-numbered PE closes. Removing just that app's override instead would arguably be worse (the env would then build the app from its default branch into the shared preview), so the conservative choice here is defensible — but it should be a documented choice, not an accident. A comment in `cleanupPreview` noting the zombie-until-last-close tradeoff would do it.

**L2 breadcrumb:** `envInUseByAnotherPreview` recomputes `fmt.Sprintf("pr-%d", ...)` — the third such site (with `:93` and the secret-watch reverse mapping). Fine for L1 since naming is unchanged, but the L2 env-rename must update all of them in lockstep; extracting a single env-name helper now would make L2 mechanical.

Issue #440, opened 2026-07-05
Harden bundled build infra: unauthenticated privileged BuildKit + writable registry, no NetworkPolicies


MC-Meesh:
The bundled build infrastructure in the umbrella chart is reachable from every tenant workload with no network isolation and no security context.

- BuildKit runs `privileged: true` and its Service (`buildkitd:1234`) accepts unauthenticated gRPC (`charts/mortise/templates/buildkit.yaml`). Any user code Mortise deploys into a `pj-*` namespace can dial it, which is effectively root on the node.
- The registry runs with `REGISTRY_STORAGE_DELETE_ENABLED=true` and no auth (`charts/mortise/templates/registry.yaml`), and the node-local proxy exposes it on `hostPort: 30500` on every node (`registry-proxy.yaml`). A tenant can push or delete images the platform then deploys (image poisoning).
- There are zero NetworkPolicies anywhere in the charts or controllers.
- The hardening in #226 added runAsNonRoot/readOnlyRootFilesystem to the operator and observer but skipped registry, registry-proxy, and buildkitd entirely (no pod or container securityContext; the `mortise-deps` namespace is PSS-labeled `privileged` across enforce/audit/warn).

**Fix:** ship NetworkPolicies restricting buildkitd/registry ingress to the operator namespace (and kubelet for the proxy); enable BuildKit mTLS and registry auth; add hardened security contexts to registry and the proxy (the Distribution registry runs fine as non-root with a read-only rootfs), and consider isolating buildkit in its own namespace. Related: #411 (app containers missing PodSecurity restricted fields) is the tenant-workload counterpart.

---
Found during a Fable 5 quality audit of the codebase (same audit that produced #433). Verified against the code at the current main.

MC-Meesh:
**Direction ruling (project steering, 2026-08-09):** infrastructure hardening moves to the user's layer. Mortise stays lightweight: its own authz model must be correct (the #439 fix, webhook HMAC, API authorization), but NetworkPolicies, BuildKit mTLS, and registry auth are cluster-operator concerns that vary too much by environment to own in the chart.

What this means for this issue:
- The **securityContext half is staying**: registry/registry-proxy/buildkit pods get values-driven security contexts (in flight on the chart-hardening lane).
[...]

PR #460, opened 2026-08-09
fix(api): gate device-flow provider creation and validate GitProvider host (SSRF)

### LINKED ISSUE #439: Device flow: unauthenticated GitProvider creation + SSRF via unvalidated host
The per-user git device flow creates cluster-scoped GitProvider CRDs with no authorization check and no host validation, yielding an SSRF and a privilege gap.

**Authz gap:** the device-flow routes (`RequestCode`/`Poll`/`StorePAT`) are mounted under `jwtAuthMiddleware` only, with no `s.authorize` call (`internal/api/server.go` ~:196-199). `getOrCreateGitProvider` (`internal/api/device_flow.go` ~:477) creates a cluster-scoped GitProvider CRD plus a webhook-secret Secret. This bypasses the admin-only gate on the intended creation path (`CreateGitProvider` in `gitproviders.go`), so any authenticated user, including a viewer, can create platform-scoped resources.

**SSRF:** `StorePAT` (`device_flow.go` ~:366-369) takes `body.Host` verbatim and passes it to `getOrCreateGitProvider`, which sets `Spec.Host` with no URL/host validation (the validation in `gitproviders.go` is not applied on this path). A later `RequestCode` makes the operator issue a server-side POST to `gp.Spec.Host + "/login/device/code"`. An attacker can point `host` at `http://169.254.169.254` or a cluster-internal service. Precondition: a default GitHub client ID is configured, which this path inherits.

**Fix:** gate on-demand provider creation behind an admin authz check, and validate `host` (scheme + public-host allowlist, pinned to the known forge hosts per provider type) before creating the provider.

---

MC-Meesh:
The per-user git device-flow endpoints (RequestCode/Poll/StorePAT) were mounted under `jwtAuthMiddleware` only, with no authorization check. Their shared `getOrCreateGitProvider` path creates a cluster-scoped GitProvider CRD on demand, so any authenticated user — including a viewer — could mint platform-scoped resources, bypassing the admin-only gate on CreateGitProvider. StorePAT additionally passed `body.Host` verbatim into `Spec.Host`; a later RequestCode issues a server-side POST to that host, yielding SSRF (e.g. `http://169.254.169.254`).

- Plumb the authz PolicyEngine into DeviceFlowHandler and gate on-demand provider creation behind an admin authz check (`gitprovider/create`), before any host input is touched.
- Extract a shared `validateGitProviderHost` validator used by both `gitproviders.go` and the device-flow path: requires an absolute http(s) URL and rejects loopback, private, and link-local hosts.
- Map the new failure modes to 403 (creation forbidden) and 400 (invalid host); genuine absence stays 404.
- Tests: viewer-role creation denial, hostile-host rejection with no GitProvider persisted, and private/link-local/loopback/non-http host rejection on the admin path.

Fixes #439

### DIFF
--- a/internal/api/device_flow.go
+++ b/internal/api/device_flow.go
@@ -5,6 +5,7 @@ import (
 	cryptorand "crypto/rand"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -22,7 +23,18 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	mortisev1alpha1 "github.com/mortise-org/mortise/api/v1alpha1"
+	"github.com/mortise-org/mortise/internal/authz"
 	"github.com/mortise-org/mortise/internal/git"
+)
+
+// Sentinel errors returned by getOrCreateGitProvider so handlers can map the
+// on-demand creation failure modes to the right HTTP status.
+var (
+	// errProviderCreateForbidden means the caller lacks permission to create a
+	// GitProvider on-demand (creation is admin-only, matching CreateGitProvider).
+	errProviderCreateForbidden = errors.New("creating a git provider requires admin privileges")
+	// errProviderHostInvalid means the caller-supplied host failed validation.
+	errProviderHostInvalid = errors.New("invalid git provider host")
 )
 
 const (
@@ -37,6 +49,7 @@ type DeviceFlowHandler struct {
 	client     client.Client
 	clientset  kubernetes.Interface
 	httpClient HTTPClient
+	authz      authz.PolicyEngine
 }
 
 // HTTPClient abstracts HTTP requests for testability.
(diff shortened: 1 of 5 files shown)


MC-Meesh (commented):

Reviewed against the mo-4jv audit evidence for #439 (lane D packet on mo-1ap). Both halves of the fix match the evidence: the authz gate sits at the top of the on-demand creation path — before any host input is touched — and host validation is a single shared boundary check used by both the admin path and the device-flow path, which is exactly the no-drift shape the packet asked for. The 403/400/404 mapping via sentinel errors keeps the existing-provider flow open to non-admins (correct: the exploit required creation, not use). Approving, with one residual to track and one test-breadth nit.

**Residual to file as a follow-up (not blocking):** `validateGitProviderHost` rejects hostile IP *literals*, but a DNS name resolving to an internal or metadata address passes (e.g. an attacker-controlled record pointing at 169.254.169.254). The authz gate reduces this to an admin-only capability, and admins can already point providers at arbitrary URLs by design — so this is defense-in-depth, not a reopened hole. Resolve-time checking has TOCTOU/rebinding limits anyway; the durable answer is egress NetworkPolicies, which is already scoped as the #440 lane. Suggest noting it on #440 so the SSRF thread has a home.

**Test-breadth nit:** viewer denial is asserted only on `/device`. All four routes share `getOrCreateGitProvider`, so the gate is structurally covered, but a route-wiring regression (e.g. a future handler bypassing the shared path) wouldn't be caught. A small table-driven loop over `/device`, `/poll`, `/token`, `/status` would pin the surface cheaply.

Two details verified while reading: `PrincipalFromContext` is populated by `jwtAuthMiddleware` on all these routes, so the nil-principal 403 is a defensive backstop rather than a reachable path; and the validator's `u.User != nil` rejection also closes the credential-smuggling shape (`https://user:pass@host`) the packet didn't explicitly list.

Issue #442, opened 2026-07-05
E2E reliability: 86 non-exact selectors + retries mask real failures


MC-Meesh:
The Playwright E2E suite has systemic selector and retry problems that launder real test failures into "flaky."

- **86 `getByRole('button', { name: ... })` calls omit `{ exact: true }`** across `ui/tests/e2e/*.spec.ts`. The repo CLAUDE.md declares this a bug, not a style choice (canvas AppNode divs have `role="button"`, so substring matching picks the wrong element and times out). Investigated CI failures were resolved precisely by adding `exact: true`/scoping.
- **`retries: 2`** in `ui/playwright.config.ts` masks deterministic selector bugs as "flaky" (a real strict-mode violation in `navigation.spec.ts` was reported as 1 flaky and would have gone green).
- **No global `expect` timeout**: the 5s default applies against an eventually-consistent backend, compensated by 200+ scattered per-assertion overrides; failures land wherever one assertion still uses the default while the controller reconciles.

**Fix:** sweep for non-exact selectors and add a CI grep/lint gate; set `retries: 1` plus fail-on-flaky (`failOnFlakyTests`); set a global `expect: { timeout: 15000 }`.

---
Found during a Fable 5 quality audit of the codebase (same audit that produced #433). Verified against the code at the current main.

MC-Meesh:
Done: #542 made every literal-named button/link role selector exact (82 sites; the remaining 9 are regex names), and #546 set Playwright retries to 0 on CI — the E2E job passed with both. Substring-matched or timing-dependent specs now fail visibly instead of passing on a retry.

PR #461, opened 2026-08-09
ci: add make ci-local — every CI gate, one command

MC-Meesh:
## Summary

Adds `make ci-local` (tracker bead mo-osm): one command that runs every gate `.github/workflows/ci.yml` runs, locally, with a per-gate pass/fail summary. Per the Overseer mandate, a green `make ci-local` is the push gate.

- Six gates mirroring the six CI jobs: unit+envtest, helm lint/template, go vet + staticcheck, UI typecheck/build, integration (k3d), UI E2E (dev cluster + Playwright).
- **Version-matched by construction**: the staticcheck version is parsed out of ci.yml's install line, so the local pin can never drift from CI's.
- **Retry parity**: the e2e gate sets `CI=true` — Playwright's config gives CI 2 retries and workers=4 but local runs 0 retries and workers=8, so an unflagged local run is both flakier and unrepresentative.
- Cheap gates run first; all gates run regardless of earlier failures (CI's jobs are independent, so a full picture per run). `SKIP_INTEGRATION=1` skips the two cluster-backed gates for the fast inner loop.
- Hardens `make test-e2e`'s port-forward with a setsid supervisor loop: `kubectl port-forward` exits on the first dropped backend connection, and a mid-suite death failed every in-flight test with `ERR_EMPTY_RESPONSE` (observed live; with the supervisor the suite then passed 160/160 three runs straight).

## Base branch note

Based on `fix/ci-hygiene-lane-f` (#458) as the bead offered, **so #458 must merge first**; this PR then sits at exactly main+1. Basing on main was tried and its integration gate fails on any containerd-store Docker host without #458's import fix — the two are functionally coupled.

## Verification

Five full runs on the crew1 box during development. Every gate has passed; the fast gates are green in every run, and ui-e2e is green in the last three runs (160/160) since the port-forward supervisor. The integration gate is red in 3 of 4 recent runs on this box due to a **product bug the gate surfaced**: App reconciles race per-namespace RBAC propagation and controller-runtime's exponential backoff turns the transient forbidden errors into minutes-long stalls, blowing test ready-ceilings. Filed with full RCA as bead mo-k5p. `ci-local` deliberately does not auto-retry that gate — hiding exactly the class of failure the gate exists to surface would defeat it.

Fixes the local-coverage gap called out when #456 hit an unexercised CI job.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,6 +316,12 @@ test-integration: ## Create k3d cluster, install chart + test deps, run integrat
 test-integration-fast: ## Run integration tests against the existing dev cluster (requires make dev-up + chart installed)
 	go test -tags integration -count=1 -timeout 5m ./test/integration/...
 
+##@ CI
+
+.PHONY: ci-local
+ci-local: ## Run every CI gate locally — THE push gate. SKIP_INTEGRATION=1 for fast mode (skips the cluster-backed gates).
+	scripts/ci-local.sh $(if $(SKIP_INTEGRATION),--skip-integration)
+
 ##@ Chart Tests
 
 .PHONY: verify-chart-dependency-drift
@@ -399,8 +405,13 @@ E2E_PASSWORD ?= admin123
 test-e2e: ## Run Playwright E2E suite against the dev cluster (requires make dev-up).
 	@kubectl --context $(DEV_KUBE_CONTEXT) -n mortise-system rollout status deployment/mortise --timeout=30s
 	@cd ui && npm ci --silent && npx playwright install chromium
-	@kubectl --context $(DEV_KUBE_CONTEXT) port-forward -n mortise-system svc/mortise $(E2E_PORT):80 >/dev/null 2>&1 & PF_PID=$$!; \
-	trap "kill $$PF_PID 2>/dev/null || true" EXIT; \
+	@# The port-forward runs under a supervisor loop: kubectl port-forward
+	@# exits on the first dropped backend connection, and a mid-suite death
+	@# fails every in-flight test with ERR_EMPTY_RESPONSE. setsid gives the
+	@# loop its own process group so the trap's group-kill reaps the loop and
+	@# whatever kubectl it currently has, without touching this recipe shell.
+	@setsid bash -c 'while true; do kubectl --context $(DEV_KUBE_CONTEXT) port-forward -n mortise-system svc/mortise $(E2E_PORT):80 >/dev/null 2>&1; sleep 1; done' & PF_PID=$$!; \
+	trap "kill -- -$$PF_PID 2>/dev/null || true" EXIT; \
 	echo "==> Waiting for API at http://localhost:$(E2E_PORT)..."; \
 	for i in $$(seq 1 30); do \
 		curl -sf http://localhost:$(E2E_PORT)/api/auth/status >/dev/null 2>&1 && break; \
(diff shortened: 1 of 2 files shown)


MC-Meesh (commented):
**Verdict: APPROVE with two should-fix nits** (posted as comment — same-account PRs cannot be formally approved).

Reviewed as a delta against the #458 branch per dispatch (the ci.yml/install.sh/run.sh hunks in this PR's file list are #458's commits riding along; the actual delta is the Makefile supervisor and `scripts/ci-local.sh`). Gate structure is right: cheap-first ordering, run-everything-collect-failures matching CI's independent jobs, `set -uo pipefail` without `-e` deliberately so one gate's failure doesn't skip the rest, and honest messaging in skip mode. On the three specific concerns:

**setsid supervisor — correct, one portability nit.** The mechanics hold up: in a non-interactive make recipe shell there's no job control, so the background child isn't a group leader, `setsid` execs without forking, and `$!` is both the loop's PID and the new process-group ID — `kill -- -$PF_PID` therefore reaps the loop *and* whichever kubectl it currently owns, which is exactly why the old single-PID trap was insufficient once a restart loop exists. Failure modes check out: cluster-down spins the loop quietly but the 30-try readiness curl below fails the recipe and the trap fires. Nit: `setsid` is util-linux — absent on stock macOS. If anyone develops on a Mac this recipe now dies with "command not found" before the suite starts; a `command -v setsid` guard with a fallback to the old single-shot behavior would keep it portable.

**staticcheck version parsing — works today, fails silently tomorrow; make it fail loudly.** `grep -oE 'staticcheck@[^ "]+' | head -1` matches the current `go install …staticcheck@v0.7.0` line fine. Two fragilities: (1) it matches the *first* `staticcheck@` anywhere in ci.yml, so a future comment or a second occurrence wins the race; (2) the `latest` fallback inverts the script's own guarantee — if the pin's syntax ever changes shape (version moved to an env var, action restored), local silently runs `latest` while CI runs the pin, which is precisely the drift this parsing exists to prevent, and the comment's claim that the fallback "keeps local == CI either way" no longer holds. Since ci-local is THE push gate, suggest anchoring the grep to the install line (`go install honnef\.co/go/tools/cmd/staticcheck@`) and **exiting with an error** when it doesn't match, rather than falling back. Two-line change.

**CI=true parity — correct for the gate's purpose, with one interplay to name.** The gate's job is to predict "would CI pass?", so mirroring `retries: process.env.CI ? 2 : 0` is right — without it, local is strictly flakier than the thing it stands in for, and the gate cries wolf. The masking this introduces is exactly #442's existing complaint about CI itself, not new masking — and when #442's config change lands (retries 1 + `failOnFlakyTests`, sequenced last in that lane), this gate tightens automatically with zero changes here, which is the payoff of keying off the config instead of hardcoding. Worth one sentence in the script comment: flake-*hunting* should run `make test-e2e` without CI=true; this gate is for parity, not diagnosis.

**One behavioral note beyond the asks:** `gate_ui_e2e` runs `make dev-down || true` unconditionally — including when the developer already had a live `make dev-up` cluster they were working against. Running the push gate currently costs you your dev environment. Consider only tearing down when `dev-up` was invoked by this script (e.g. detect cluster existence first), or at minimum saying so in the usage header.

Merge-train: this PR contains #458's commits, so it must land after #458 (or rebase if #458 is amended). Known flake exposure: the integration gate inherits mo-k5p (RBAC-propagation stalls, ~75% local failure rate on the crew1 box) — that's a pre-existing product flake this gate correctly *surfaces* rather than hides, per its charter; triage for it is going to the mayor separately.

Issue #379, opened 2026-05-15
Stale mortise-core subchart tarball missing BuildRun RBAC


JamesSprow:
## Problem

The umbrella chart at `charts/mortise/charts/mortise-core-0.1.0.tgz` is a stale packaged copy of the `mortise-core` subchart. It was never repackaged after BuildRun RBAC rules were added to `charts/mortise-core/templates/rbac.yaml`.

The `Chart.yaml` dependency points to `file://../mortise-core`, but Helm uses the existing `.tgz` over the local path reference. As a result, `helm upgrade` deploys a ClusterRole missing the `buildruns`, `buildruns/status`, and `buildruns/finalizers` resources, causing the operator to fail with:

```
failed to list *v1alpha1.BuildRun: buildruns.mortise.mortise.dev is forbidden:
User "system:serviceaccount:mortise-system:mortise-controller" cannot list
resource "buildruns" in API group "mortise.mortise.dev" at the cluster scope
```

## Stale tarball RBAC (missing buildruns)

```yaml
- apiGroups: ["mortise.mortise.dev"]
  resources: ["apps", "apps/status", "apps/finalizers"]
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```

## Source template RBAC (correct)

```yaml
- apiGroups: ["mortise.mortise.dev"]
  resources: ["apps", "buildruns"]
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
- apiGroups: ["mortise.mortise.dev"]
  resources: ["apps/finalizers", "buildruns/finalizers"]
  verbs: ["update"]
- apiGroups: ["mortise.mortise.dev"]
  resources: ["apps/status", "buildruns/status", ...]
  verbs: ["get", "update", "patch"]
```

## Fix

Run `helm dep update charts/mortise` to repackage `mortise-core-0.1.0.tgz` from the local source, and commit the updated tarball.

Consider adding a CI check or Makefile target that verifies the packaged subchart is up to date with the source.

JamesSprow:
Regression observed on 2026-07-24 with the public `mortise/mortise` chart version `0.1.0`:

```sh
helm upgrade mortise mortise/mortise -n mortise-system --version 0.1.0 --reuse-values ...
```

The resulting `mortise-controller` ClusterRole again lacked `buildruns`, `buildruns/status`, and `buildruns/finalizers`. After the controller restarted it repeatedly logged:

```text
[...]

PR #462, opened 2026-08-10
release: verify the published chart artifact; fix latent chart-test and installer bugs

### LINKED ISSUE #446: helm upgrade can remove registry-proxy despite registry.enabled=true

### LINKED ISSUE #454: install.sh still sets renamed metricsServer.enabled key — script silently no longer disables metrics-server

MC-Meesh:
## Summary

Lane F2 (tracker bead mo-5yb): make the published chart artifact trustworthy, add the #446 regression lifecycle test, and fix the #454 installer key — plus three latent bugs surfaced by running the chart-integration suite for what turned out to be the first time anywhere.

**Publish-layer verification** (`release.yml`): after publishing, the workflow pulls the chart back through the public gh-pages repo URL (polling for Pages deployment), asserts it is **byte-identical** to the tarball the run packaged, renders the pulled copy, and asserts the registry-proxy DaemonSet (`registry.enabled=true`) and buildruns RBAC entries are present — the two things whose absence drove #446 and #379's stale-template reports. Publishing now also **refuses to overwrite an already-published version** (chart artifacts are immutable) and refuses to publish the placeholder version. This is the published-artifact counterpart of `make verify-chart-dependency-drift`, which guards the in-repo vendored subchart.

**Placeholder version policy**: the in-repo charts now carry `0.0.0-dev` instead of `0.1.0`, which collided with the genuinely published `mortise-0.1.0` release artifact. Documented in RELEASING.md; CI stamps real versions at release time exactly as before (the drift check derives the tarball name from Chart.yaml, so nothing else moves).

**#446 regression test** (`test/chart/run.sh` Test 3b): registry-proxy present after fresh install → **still present after a bare `--reuse-values` upgrade** (the exact upgrade shape that lost it) → gone on `registry.enabled=false` → back on re-enable. Uses the #458 `HELM_WAIT_TIMEOUT` constant.

**#454**: `install.sh` passed `--set metricsServer.enabled=false`, a key renamed to `metrics-server` in 1.0.4 that Helm silently ignores — metrics-server deployed anyway. Fixed (quoted hyphenated key), docs table updated, and a CI grep now fails on the stale spelling outside CHANGELOG (bracketed pattern so it can't match itself).

## Latent bugs found by actually running the suite

`make test-chart-integration` is referenced as "[release only]" but is wired into **no workflow** — it had never executed anywhere (tracked as bead mo-jp7 for wiring). Its first real runs surfaced, and this PR fixes:

1. `run.sh` built the operator image with an **untargeted `docker build`** — the Dockerfile's final stage is the **observer** binary, so the operator deployment crashlooped and the install timed out.
2. `scripts/install.sh` has the identical bug in its repo-clone dev path — **user-facing**: the installed operator crashloops.
3. `run.sh`'s traefik readiness check queried the release namespace; the chart deploys traefik into the deps namespace, so the check could never pass.

## Verification

- `make ci-local` fast gates green; full run green except the integration gate's known mo-k5p RBAC-race flake (single test at its ready-ceiling, forbidden-error signature confirmed in operator logs, occurrence recorded on the bead — this branch touches no controller code).
- `make test-chart-integration`: **13/13 PASS locally** — first green run in the suite's existence, including all four new Test 3b assertions and the full install-script path (which now also exercises the fixed metrics-server flag for real).
- release.yml verification assertions validated against a local render before shipping; both workflow files YAML-validated.

Fixes #446. Fixes #454. Completes the #379 publish-layer residual (#379 itself is closed).

## Coordination

- Builds on #458's `HELM_WAIT_TIMEOUT`; no literals reintroduced.
- Complementary to the in-repo drift check per the lane packet — naming and comments cross-reference the two as one system.
- `run.sh` was otherwise unclaimed after #458; lane E untouched.


MC-Meesh (commented):
**Verdict: APPROVE with one doc-structure fix requested** (posted as comment — same-account PRs cannot be formally approved).

Reviewed against the mo-4jv audit's #446/#379 publish-drift picture and the lane spec. This lands the whole thesis: the published artifact — not the source tree — is what users install, and now both the placeholder collision (the root cause that made "#446 as filed" unreproducible from source) and the publish path itself are guarded. Several details are better than the spec asked for.

**Verified off-diff, all three latent-bug fixes are real:**
- The Dockerfile's stage order is `ui-builder → builder → operator → observer` — the final stage is genuinely the **observer**, so an untargeted `docker build` tags the observer binary as the operator image and the deployment crashloops. `--target operator` is load-bearing in both `run.sh` and `install.sh`, and the installer one is user-facing (the repo-clone dev path shipped broken). The why-comments are right.
- `charts/mortise/values.yaml` sets `traefik.namespaceOverride: mortise-deps` — the readiness check was querying the release namespace and could never find it. Fix is correct. **Follow-up this exposes: ARCHITECTURE.md is now known-stale** — its component table and diagram place Traefik in `mortise-system`. Trivial doc fix, separate PR or ride-along.
- The `verify-chart-dependency-drift` target derives the tarball name from `charts/mortise-core/Chart.yaml` (`awk '/^version:/'`), so the `0.0.0-dev` rename is compatible — checked, no hardcoded `0.1.0` left behind.

**Release pipeline:** the placeholder-refusal guard, the immutability check (nicely done: `cmp -s` allows byte-identical republish, so a failed workflow can re-run idempotently while a content change is refused), and the post-publish pull-back (poll index → `helm pull` through the public URL → byte-compare → render → assert registry-proxy DaemonSet + buildruns RBAC) exactly match the mechanism the audit identified behind both #446 and #379's regression report. The `0.0.0-dev` placeholder choice is the right one — an implausible version makes "dev tree or release?" answerable at a glance.

**Test 3b** is the requested #446 lifecycle verbatim, including the bare `--reuse-values` upgrade with no `--set` flags — exactly the shape the report described — and it uses the #458 `HELM_WAIT_TIMEOUT` var rather than reintroducing literals.

**Requested fix (small): RELEASING.md section placement.** The new `## Placeholder version policy` H2 is inserted *between* the `## Hotfix / rollback` heading and its body — "If a release is broken:" and the rollback instructions now render under the placeholder-policy heading. Move the new section above `## Hotfix / rollback` (or below its body).

**Two non-blocking notes:**
1. Verification runs *after* the gh-pages push, so a verification failure leaves the bad artifact published (workflow goes red and the Release job is blocked, but a pinned `--version` install is served the artifact until someone acts). A cheap pre-push `helm template` assert on the packaged tarball would catch template regressions before publishing; keep the post-publish check as the end-to-end proof of the publish path itself. Worth a follow-up.
2. Tiny: in the verify step, the standalone `grep -q buildruns` is subsumed by the `buildruns/finalizers` and `buildruns/status` checks, and the `-B5`-based DaemonSet grep depends on rendered field ordering — both fine in practice, just brittle enough to mention.

The `metricsServ[e]r` CI guard (bracket trick to avoid self-matching, CHANGELOG excluded) plus the docs/install.md key fix close #454 completely.

Issue #351, opened 2026-05-12
Shared vars bypass restricted-environment secrecy for project developers


JamesSprow:
## Summary
Project-level shared vars remain plaintext-readable and writable for project developers even when those vars are materialized into restricted environments.

## Current behavior
- `GET /api/projects/{project}/shared-vars` returns plaintext shared values for a project developer.
- `PUT /api/projects/{project}/shared-vars` also allows the same developer to update those values.
- Those shared vars are then materialized into restricted environments, so a developer without update access to the restricted env can still read or modify values that become effective there.

## Why this is a problem
The env redaction work correctly gates per-environment plaintext access on effective env-scoped update permission, but shared vars still use only project-level update permission. That makes restricted-environment secrecy inconsistent depending on whether a secret is stored directly on the env or inherited through shared vars.

## Repro outline
1. Create a project with a restricted environment such as `production` and a non-restricted environment such as `staging`.
2. Add a project developer who lacks update access to the restricted environment.
3. Store a sensitive value in shared vars.
4. Call `GET /api/projects/{project}/shared-vars` as that developer.
5. Observe the plaintext value is returned even though the value is effective inside the restricted environment.

## Notes
This likely needs a design decision across both read and write paths, not just a small redaction patch.

MC-Meesh:
Closing as **won't-fix — working as designed** (project steering ruling, 2026-08-09).

Shared variables are **project-visible by definition**: they exist precisely to be the project-wide layer that every environment inherits. The audit confirmed the mechanics this issue describes — `GetSharedVars`/`PutSharedVars` authorize at project scope with no environment dimension (`internal/api/env.go`), so a project Developer can read and write them even when a restricted environment consumes them. The ruling is that this is the intended contract, not a leak: **anything a Developer must not see or touch belongs in a per-environment variable on the restricted environment**, where the restricted-env authorization branch (`internal/authz/native.go`) fully applies. Splitting shared vars into ACL tiers was considered and rejected as machinery v1 doesn't need.

Documentation follow-up: the restricted-environments section will state this boundary explicitly ("shared vars are project-visible; use env-scoped vars for restricted secrets").

One verification task remains open internally (not this issue): confirming shared vars reliably propagate into newly created environments and preview environments, so the contract that justifies this ruling actually holds everywhere.

PR #463, opened 2026-08-10
test: prove the shared-vars propagation contract (#351 ruling follow-up)

MC-Meesh:
## Summary

Verification lane for the #351 Overseer ruling (tracker bead mo-3gd): shared vars stay as implemented, and this PR delivers the one remaining obligation — **proof** that they reliably reach newly spawned environments.

**Outcome: the contract holds; no fix was needed.** The audit's suspicion was that a non-clone new environment could start without shared vars (the API create path copies no secrets and pokes no apps). Code-reading plus live tests show the machinery covers it: any env-set change on the Project makes the project controller's `cascadeEnvChange` bump `ProjectEnvsRevAnnotation` on every App, and each app reconcile re-materializes `shared-env` in every resolved env namespace from the control-namespace source of truth — so ordering can't lose writes, restricted envs are included (`resolveEnvs` doesn't filter them), and previews participate both at clone time and in every later fan-out.

Three integration tests pin this permanently:

| Test | Contract point |
|---|---|
| `TestSharedVarsReachNewEnv` | var exists → bare spec-append env (non-clone API shape) → Secret materializes with **no poke** |
| `TestSharedVarsReachRestrictedEnv` | same guarantee, `restricted: true` |
| `TestSharedVarsUpdateConvergesLiveEnvs` | update + PutSharedVars-shaped poke → prod **and** live preview env converge |

## Verification

Full `make test-integration`: **56/56 PASS** including the three new tests (26s / 28s / 24s). `make ci-local` fast gates green.

Completes the verification obligation from the #351 wont-fix ruling (#351 itself stays closed).


MC-Meesh (commented):

Reviewed against the mo-3gd contract spec, with the specific question "would these tests catch the gap if it existed?" — yes, and I independently verified the no-product-change conclusion.

**The conclusion is correct; the audit's suspicion is disproven.** The audit hypothesized that a non-clone env addition might never trigger app fan-out. Verified in main: `envsChanged` compares the spec env-name list against `Status.Environments` (`project_controller.go:262-273`), so an *addition* trips it; `cascadeEnvChange` (`:276-296`) then bumps `ProjectEnvsRevAnnotation` (rev = `project.Generation`, idempotent via the equal-rev skip) on every App in the control namespace, and the App reconcile re-materializes the shared-env Secret from the control-ns source. The mechanism lives in the project controller — a path the audit's fan-out analysis didn't credit. Happy to be wrong; this is exactly what the verification bead existed to settle.

**The tests genuinely pin the contract.** `TestSharedVarsReachNewEnv` is the suspected-gap scenario reproduced faithfully: `appendProjectEnv` is a bare `RetryOnConflict` spec append — the same shape as the API's `createProjectEnvironment` — with explicitly no poke and no secret copy. If the cascade didn't exist, nothing would ever materialize the Secret in the new namespace and the 2-minute `RequireEventually` would fail. The restricted variant covers the ruling's explicit scope, and the convergence test verifies both the poke path (the `mortise.dev/env-updated` key matches `internal/api/env.go:653` exactly) and preview participation in the fan-out; the comment's deferral of creation-time preview inheritance to `TestPreviewEnvironmentInheritsSourceEnvVars` checks out — that test exists (`env_clone_test.go:31`).

**One operationally important note: these three tests join the mo-k5p flake population.** They create fresh projects, wait for apps in brand-new namespaces (5-minute `WaitForAppReady` ceilings), and the cascade-triggered reconciles land in namespaces whose RoleBindings may still be propagating — precisely the RBAC race mo-k5p describes, at ~75% local suite failure on the crew1 box. Until mo-k5p's fix lands, a timeout in these tests is far more likely to be that race than a contract violation. Worth one sentence in the file's header comment so a future flake isn't misread as the contract breaking.

Nits, non-blocking: (1) `pokeAllApps` hardcodes `"mortise.dev/env-updated"` — faithful to the product code, which also uses a literal; if that key ever moves to `internal/constants` (REVIEW.md discipline), move both together. (2) Restricted-env coverage is creation-only; update-convergence is tested for prod+preview but not restricted — the mechanism is env-agnostic so this is optional completeness, cheap to add if the file is ever touched again. (3) Asserting the env-ns Secret is the right contract level (pod-level injection is covered elsewhere by the binding/e2e suites).

On merge, mo-3gd can close with this PR as its evidence — verification outcome: contract holds, no product change needed.

Issue #444, opened 2026-07-05
Audit roll-up: lower-priority auth, controller, and chart RBAC hardening


MC-Meesh:
Roll-up of lower-severity hardening items from the audit. Each is small; grouped to avoid issue sprawl. Split out if any gets picked up.

**Auth**
- No login throttling or lockout on `/api/auth/login`, plus a username-enumeration timing oracle: `NativeAuthProvider.Authenticate` returns immediately on user-not-found without running bcrypt, while a found user always pays the bcrypt cost (`internal/auth/native.go` ~:51-70). Add a dummy bcrypt compare on the not-found path and per-IP/per-account rate limiting.
- Apps with an empty `providerRef` are redeployed by any provider signed webhook whose repo string matches (`internal/webhook/handler.go` ~:166,219). Treat empty as no-match, or bind apps to a specific provider.
- Webhook returns 404 for unknown provider vs 401 for bad signature, letting an unauthenticated caller enumerate configured provider names (`handler.go` ~:82-116). Return a uniform status.
- JWT refresh accepts access tokens expired up to 7 days (`internal/auth/jwt.go` ~:96-110). Mitigated by live `pwd_gen` checks; consider a dedicated refresh token or shorter leeway. (See also #282.)

**Controllers**
- `time.Now()` used directly in `internal/controller/project_controller.go` ~:722 (ProjectReconciler has no injected `clock.Clock`), which CLAUDE.md declares a bug. Add a Clock field.
- Registry PushTarget/PullTarget errors are swallowed and the env is silently skipped with no status condition (`app_controller.go` ~:780-789); a persistent registry misconfig leaves the app `Ready` on a stale image with the only signal in operator logs. Surface as a condition or return the error.
- The build-status flush copies the whole in-memory `Environments` slice over a freshly-read status (`app_controller.go` ~:336-346), which can transiently stomp a concurrent BuildRun status write. Merge per-env build fields instead of wholesale replacement.

**Chart RBAC**
- The chart ClusterRole is missing `gitproviders/finalizers`, `platformconfigs/finalizers`, `previewenvironments/finalizers` (drift vs the generated role; bites under OwnerReferencesPermissionEnforcement).
- The static RoleBinding in the operator namespace binds the full `mortise-controller-ns` write role where the stated intent is OAuth/config secret access only; a narrow Role would match intent.

---
Found during a Fable 5 quality audit of the codebase (same audit that produced #433). Verified against the code at the current main.

[...]

PR #464, opened 2026-08-10
fix: classify RBAC-propagation races as fast requeues across three controllers

MC-Meesh:
## Summary

Fixes the mo-k5p P1: reconciles racing per-namespace RBAC bootstrap saw Forbidden errors that fed controller-runtime's exponential rate limiter (max 1000s), compounding a seconds-long propagation window into minutes-long stalls — red-ing the local integration gate in ~75% of full runs. Operator-log forensics across four instrumented suite runs revealed the race has **three phases**, and all three are now classified as errorless 2-second requeues, bounded to namespaces younger than 2 minutes (persistent Forbidden in an older namespace = genuine misconfiguration → `Failed` with an `RBACForbidden` condition that clears on recovery):

1. **Forbidden in a young existing env namespace** — the classifier (originating in chrome's implementation, kept with its five unit tests) is now wired at **all eleven** env-write sites in the App controller, including the deployment/cronjob paths that carry the env-Secret writes (the dominant denial in the logs) and the external-source loop.
2. **Forbidden in a namespace that doesn't exist yet** — the authorizer denies before existence is checked, so writes racing namespace creation itself (a preview env override landing on an App before the Project controller creates `pr-N`) surfaced as Forbidden-in-missing-namespace and slipped past the ns-age check. Run-2 forensics: 42 raw errors vs 2 classified. Now the earliest-phase fast requeue.
3. **The Project controller held every env RoleBinding hostage to its own race** — its control-ns bookkeeping (owner member, create activity) ran *before* the env-namespace loop, so one Forbidden there aborted the reconcile into rate-limited backoff and delayed env RoleBinding creation for minutes. The bookkeeping now runs after env namespaces + RoleBindings exist, with its own young-ns Forbidden fast-requeued. The PreviewEnvironment controller's secret copies get the same classification.

## Verification

- Unit: classification tested directly (young / old / boundary / **missing-ns** / non-forbidden / **wrapped**) — envtest runs the operator as cluster-admin, so RBAC denial can't be integration-tested.
- Suite-level proof (the flake-rate delta): baseline **4 of 6 recent full `make ci-local` runs red** on this signature (rotating test subsets, 31–65 forbidden errors per failing run). Post-fix: **two consecutive integration-gate greens at 465s and 243s** — the 243s run is the fastest suite time recorded on this box, consistent with zero backoff stalls. The one red gate in run 4 was `deployments.spec.ts` e2e — the separately tracked mo-kcl 409-retry flake, occurrence recorded there.

## Notes

- Builds on the auto-saved implementation from polecat/chrome/mo-k5p+msmr0875 (classifier + tests kept; wiring completed; two more phases discovered and fixed).
- Scope deliberately excludes crew2's fix-direction (b) (SSAR-before-env-ready) — phase-2 hardening, separate bead.
- No overlap with the #437 adoption-guards bead's function bodies beyond the already-coordinated call-site wraps.

Fixes the tracker bead mo-k5p (P1).


MC-Meesh (commented):
**Verdict: APPROVE, with one substantive should-fix (phase C is unbounded) tracked as a follow-up bead** (posted as comment — same-account PRs cannot be formally approved).

Reviewed against the mo-k5p triage spec. The classifier matches the triage exactly where the triage applied (Forbidden + young namespace → errorless 2s requeue that starves the rate limiter; boundary treated as old and pinned by a test; old namespace → Failed + `RBACForbidden` condition + real error). The two discoveries beyond the triage — the missing-namespace phase and the project controller's self-inflicted backoff coupling — are both real and both correctly diagnosed. Answers to the three review questions:

**1. Bounded-window semantics — phases A and B are right; phase C escapes the bound.** The missing-namespace branch (`forbiddenFastRequeue` NotFound arm, and its twin in `envResourceError`) fast-requeues forever with no error, no condition, and only Info-level logs. The comment says "the namespace is on its way" — but nothing enforces that. Concrete scenario: the project controller cannot create the env namespace at all (RBAC to create namespaces revoked, quota, or a crashlooping project controller) → the App controller polls at 0.5 Hz indefinitely, silently, and the App never reaches Failed. This is exactly the "genuinely-misconfigured setup retries forever" case the bounded window exists to prevent — the bound is keyed to namespace age, and a namespace that never exists has no age. Mitigating factors, honestly stated: in that scenario the *project* controller is erroring loudly on its own reconcile, so the system isn't signal-free — but the App-side state is a silent masking loop, and an operator looking at the App sees nothing. **Cheap bounded design that needs no new state:** on the first phase-C hit, set a `NamespacePending` condition (timestamped by `SetStatusCondition`'s `LastTransitionTime`); on subsequent phase-C hits, if `Since(LastTransitionTime) > rbacPropagationWindow` (or a slightly larger namespace-creation budget), escalate to Failed exactly like phase B; clear it alongside `rbacForbiddenCleared`. Filed as a follow-up bead rather than blocking, since the unbounded state requires an already-loudly-failing project controller — but it should land soon after.

**2. Reorder side effects — one real coupling reversal to accept knowingly.** Moving `ensureOwnerMember`/`ensureProjectCreateActivity` after the env-namespace loop fixes the hostage problem and both calls are idempotent, so the move is safe in the healthy path. But the failure coupling now runs the other way: previously a broken env-namespace creation still left the creator with their owner ProjectMember; now it blocks membership creation entirely — a non-admin creator whose project hits an env-ns bootstrap failure can't see their own project (member-scoped authz) until that's resolved. Narrow, loud-failure-adjacent, and probably acceptable — but it's a behavior change that deserves a sentence in the reorder comment. The alternative that preserves both properties (keep the original order, but on a bookkeeping Forbidden *continue* to the env loop and requeue at the end) costs more complexity; fine to reject it, just do it on purpose. Activity-entry delay is cosmetic. Relative order vs `gcStaleEnvNamespaces`/`cascadeEnvChange`/`markReady` is preserved.

**3. 11/11 coverage — verified, and the uncounted writers are safe for the right reasons.** Main loop: PVCs, ConfigMaps, ServiceAccount, CredentialsSecret, CronJob, Deployment, Service, Ingress (8); external-source path: CredentialsSecret, ExternalNameService, Ingress (3). All eleven route through `envResourceError`. Writers deliberately not wrapped, checked one by one: BuildRun creation targets the control namespace under *cluster-scoped* CRD RBAC (no propagation race); env-var Secret materialization surfaces through the wrapped Deployment/CronJob call sites; `gcOptedOutEnvs`/finalizer GC target long-lived namespaces outside the race window. PE controller: both env-namespace write sites wrapped, and the classifier keying on the *preview* namespace is correct for the read-from-source case too (a source-ns read denial in an old source ns correctly falls through to a real error). Project controller: both control-ns bookkeeping writes wrapped.

**Smaller observations, no action required:** (a) the `rbacForbiddenCleared` clear can fire on a pass where a build-requeue `continue` skipped an env's writes entirely — the condition would flap off and back on rather than staying wrongly cleared, so it's cosmetic; (b) the wrapped-forbidden table case usefully pins that this apimachinery version unwraps in `IsForbidden` — keep that test forever; (c) freebie for the #444-controllers bead: `ProjectReconciler` now has the injectable `Clock` that its `time.Now()` item needs — wiring it there just got trivial.

The two consecutive integration greens with a fastest-ever wall clock are consistent with the diagnosis: the fix removes minutes of compounding backoff from the exact tests that were timing out.

MC-Meesh:
**Correction to my review's coverage verification.** I wrote that env-var Secret materialization "surfaces through the wrapped Deployment/CronJob call sites" — true for the deploy path, but there is a **12th call site** the sweep and my check both missed: `reconcileResolvedEnvSecrets` (`app_controller.go:1493`), invoked from the git-source *preflight* path (`:231`) so bindings/shared vars materialize even when `prepareGitSource` says don't-proceed. Its error returns as a plain error without routing through `envResourceError` — forbidden in a fresh preview/staging namespace feeds the rate limiter, which is the original k5p compounding.

Caught empirically while running the integration gate for the mo-p2l lane on a post-#464 tree: 31 forbidden errors with the signature `reconcile env secret for env pr-100: materialize shared-env: … secrets is forbidden`, and the same test names timing out (`TestPreviewInheritsStagingBindings`, `TestFromBindingProjectsNamedKey`). Tracked with evidence and a proposed S fix as bead mo-tpe. The 11 wrapped sites are working as reviewed — the fast-requeue classifier is doing its job everywhere it was installed.

PR #465, opened 2026-08-10
fix(api): compute envstore mutations inside the conflict-retry callback

### LINKED ISSUE #415: Concurrent env var PATCH silently loses data (read-modify-write race)
## Summary

Concurrent PATCH requests to the env var endpoint silently lose data. The handler pre-computes the desired Secret data from a stale read, then uses `UpdateWithConflictRetry` which retries the write but still applies the stale pre-computed data, clobbering changes from other concurrent requests.

## Reproduction

```bash
# 5 concurrent PATCHes — each sets one unique var
for i in $(seq 1 5); do
  curl -X PATCH "/api/projects/portfolio/apps/portfolio/env?environment=production" \
    -d "{\"set\":{\"RACE_$i\":\"$i\"},\"unset\":[]}" &
done
wait
```

**Expected:** All 5 vars present (RACE_1 through RACE_5)  
**Actual:** Only 1 var survives (the last writer). 2 requests get conflict errors, 2 silently lose their data despite reporting `{"status":"updated"}`.

## Root cause

In `PatchEnv()` (env.go:171-287):
1. Reads existing vars via `store.Get()` (line 212)
2. Computes final `result` slice in memory (lines 218-262)
3. Calls `store.Set()` → `upsertSecret()` → `UpdateWithConflictRetry()` (line 274)

In `UpdateWithConflictRetry` (retry.go:15-45):
- Re-fetches the Secret on each retry (line 24)
- But the `mutate` callback (envstore.go:394-398) **replaces** `existing.Data` with the pre-computed `desired.Data`
- The pre-computed `desired.Data` was built from the stale read in step 1
- On retry, the fresh Secret data (which may contain other requests' writes) is **clobbered**

## Impact

- Silent data loss on concurrent env var operations
- Particularly dangerous during CI/CD where multiple services might update env vars simultaneously
- The "updated" response is misleading — the write succeeded at the k8s level but overwrote concurrent changes

## Fix

Move the read-compute-write cycle inside the `UpdateWithConflictRetry` callback so that set/unset operations are applied to the freshly-read Secret on each attempt.

MC-Meesh:
## Summary

Lane G (tracker bead mo-gmu): eliminates the lost-update race on concurrent env-var writes. Every envstore mutation used to compute its desired result from a read taken **before** the conflict-retry loop — the retry's fresh Get bought only a fresh resourceVersion, so a retry after a conflict wholesale-replaced the Secret with the stale precomputed value, silently dropping the very write it had conflicted with.

- **`Store.Apply`**: a mutate-callback API generalizing the one call site that already had the correct read-modify-write idiom (`Delete`). The callback receives current vars freshly read on **every** attempt (nil when the Secret doesn't exist) and returns the full desired set. `Set`/`Merge`/`MergeShared`/`MergeSharedSource`/`ReplaceSource` are reimplemented on it; a lost create race surfaces as a conflict so the next attempt applies as an update; `ErrSkip` declines to create an empty Secret.
- **Retry policy**: `retry.DefaultRetry` (5 attempts, jittered backoff) replaces 3 attempts with none.
- **Call sites**: `PatchEnv` and `ImportEnv` pass their merge logic as the callback (envstore conflict errors map to HTTP responses); the App controller's `reconcileEnvSecret` — previously four separate racy writes (stale-spec deletes, seed merge, binding-source replace) — is now one recomputed-in-callback Apply, so a reconcile can't clobber a concurrent user PATCH or vice versa. Clone and shared-vars-replace sites keep `Set` semantics deliberately: their results don't depend on the target's current state, and `Set` itself is now conflict-safe.
- **Concurrency tests** (none existed): interceptor-injected conflicts prove two concurrent `Apply` writers both survive, a lost create race retries as an update, `ErrSkip` creates nothing, and a handler-level concurrent-PATCH test asserts no dropped keys.

## Provenance

Implementation and tests are polecat rust's (auto-saved branch `polecat/rust/mo-gmu+msmtf4wq` after a session death, pre-verification). Evaluated against the lane spec: complete and correct — every real RMW call site converted, test shapes exactly as specified. This PR is that work rebased onto current main, squashed, and verified.

## Verification

- `internal/envstore`, `internal/api`, `internal/controller` test suites green; `make ci-local` fast gates green.
- Full `make ci-local`: 5/6 green; the integration gate failed on exactly the known mo-k5p signature (`TestFromBindingProjectsNamedKey` at its ready-ceiling; 64 forbidden errors, **zero** fast-requeue lines — as expected, since this branch predates #464's classifier). This is a useful control: main-without-#464 still flakes, corroborating #464's before/after delta. Zero overlap with this diff (different failure plane entirely).

Fixes #415.

### DIFF
--- a/internal/api/env.go
+++ b/internal/api/env.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -40,6 +41,23 @@ type envVarResponse struct {
 type patchEnvRequest struct {
 	Set   map[string]string `json:"set"`
 	Unset []string          `json:"unset"`
+}
+
+// envConflictError is returned from envstore.Apply callbacks when a request
+// tries to modify a managed (non-user) var. Maps to HTTP 409.
+type envConflictError struct{ msg string }
+
+func (e *envConflictError) Error() string { return e.msg }
+
+// writeEnvStoreError maps an envstore.Apply error to the HTTP response:
+// envConflictError → 409, everything else → the generic error mapping.
+func writeEnvStoreError(w http.ResponseWriter, r *http.Request, err error) {
+	var conflict *envConflictError
+	if stderrors.As(err, &conflict) {
+		writeJSON(w, http.StatusConflict, errorResponse{conflict.msg})
+		return
+	}
+	writeError(w, r, err)
 }
 
 func declarativeEnvSources(app *mortisev1alpha1.App, envName string) map[string]string {
@@ -178,44 +196,6 @@ func (s *Server) PutEnv(w http.ResponseWriter, r *http.Request) {
 	}
 	store := &envstore.Store{Client: s.client}
 
-	// Read existing vars so we can preserve non-user sources (binding, generated, shared).
-	existing, err := store.Get(r.Context(), envNs, app.Name)
-	if err != nil {
-		writeError(w, r, err)
-		return
-	}
-	existingSource := make(map[string]string, len(existing))
-	for _, e := range existing {
-		existingSource[e.Name] = e.Source
-	}
-	declarativeSource := declarativeEnvSources(app, envName)
-	for _, v := range vars {
-		if source, ok := declarativeSource[v.Name]; ok {
-			writeJSON(w, http.StatusConflict, errorResponse{fmt.Sprintf(
-				"cannot overwrite %q: managed by %s (only user-set vars can be modified)", v.Name, source)})
-			return
-		}
-		if src, ok := existingSource[v.Name]; ok && src != "" && src != "user" {
-			writeJSON(w, http.StatusConflict, errorResponse{fmt.Sprintf(
-				"cannot overwrite %q: managed by %s (only user-set vars can be modified)", v.Name, src)})
-			return
-		}
-	}
-
-	// Start with non-user vars and declarative projections from the existing
-	// Secret. secretRef values resolve with source "user", so the App spec is
-	// the ownership signal that distinguishes them from editable literals.
-	var merged []envstore.Env
-	for _, e := range existing {
-		if (e.Source != "user" && e.Source != "") || declarativeSource[e.Name] != "" {
-			merged = append(merged, e)
-		}
-	}
-	// Append the new user vars.
-	for _, v := range vars {
-		merged = append(merged, envstore.Env{Name: v.Name, Value: v.Value, Source: "user"})
-	}
-
 	projectName, ok2 := constants.ProjectFromControlNs(app.Namespace)
 	if !ok2 {
 		writeJSON(w, http.StatusInternalServerError, errorResponse{fmt.Sprintf("app %q not in a control namespace (%q)", app.Name, app.Namespace)})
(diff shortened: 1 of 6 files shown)


MC-Meesh (commented):

Reviewed against the #415 audit and the mo-gmu lane spec. Crew1's complete-to-spec assessment holds — this is the strongest shape of the fix: `Apply` generalizes the `Delete` idiom exactly as specified, all five primitives now compute inside the retry, and the create-race-as-conflict conversion (surface `AlreadyExists` as a synthetic `Conflict` so `RetryOnConflict` re-reads and applies as an update) is an elegant closure of the one gap the audit called out in the old Create branch. Answers to the three review questions:

**1. The wider-than-filed controller race is covered.** The audit's concern was that `reconcileEnvSecret`'s read → stale-delete → `Merge` → `ReplaceSource` sequence — four separate writes, each computing from the initial read — let the reconciler clobber a concurrent user PATCH. The collapse into one `Apply` closes it structurally: stale-spec removal, seeding, and binding-source replacement all recompute from `current` on every attempt, so a racing PATCH forces a conflict, the retry re-reads, and the user's key rides through the `kept` logic (a user var with a fresh value never matches the `tracked && value == lastVal` removal predicate). Hoisting spec-var *resolution* outside the callback is correct — those reads target other Secrets, not the one being mutated, and their values don't participate in the conflict. Two behavior notes, both fine: the `deleteBindingOnlyEnvSecret` early-return is gone, so a binding-only Secret now gets deleted and immediately recreated with spec/kept content by the `Apply` — net semantics equivalent, and the transient-absence window is pre-existing class, not new; and result ordering changed, which is inert since `buildSecret` maps by name with bindings appended last preserving the collision-wins rule.

**2. The two kept `Set` sites — agree, with the criterion stated so it stays true.** Wholesale-replace is correct exactly where (a) the operation's *contract* is "replace the entire set" and (b) no partial writer shares the Secret. Both survivors qualify: the clone path's intent is "make target a copy of source" (a concurrent PATCH losing to a clone is the desired outcome), and the shared-vars full-PUT path is a whole-set replacement API whose only co-writer, `MergeSharedSource`, is now Apply-based — so the last partial writer that could have been clobbered no longer exists. That second condition is an *invariant*, not a property of `Set` itself: a future partial writer of either Secret would silently resurrect the race. Suggest one doc-comment line on `Set`/`SetShared`: "replace-semantics only — partial updates must use Apply/Merge." Cheap insurance on a subtle argument.

**3. #351 coordination — confirmed dissolved, nothing resurrected.** The diff touches `PutEnv`/`PatchEnv` and the store internals; `GetSharedVars`/`PutSharedVars` authorization semantics are untouched, so nothing re-enters the territory the wont-fix ruling closed. Also checked the interaction with #463's just-merged contract tests: `MergeSharedSource` keeps its signature and gains conflict-safety, so those tests are unaffected (mildly strengthened).

**Beyond the three questions:** the `retry.go` rewrite onto `retry.DefaultRetry` delivers the "widen retries + backoff" item from the lane spec (5 jittered attempts vs. 3 dry ones) while preserving the old exhaustion and no-change semantics; `validateEnvVars` moving inside `applySecret` centralizes validation over every path including the callback results; `envConflictError` propagates un-wrapped through `RetryOnConflict` (not Conflict-shaped, so no spurious retries) and maps cleanly to 409. The test suite matches the spec'd interceptor pattern — the mid-flight competing-Apply test is precisely the lost-update reproduction the audit asked for, and the 6-writer handler-level test pins the contract at the HTTP boundary. The integration flake being a pure pre-#464 k5p signature is consistent with everything reviewed here; no interaction between this diff and the RBAC classifier paths.

PR #466, opened 2026-08-10
fix(controller): ownership guards on six workload adoption/update/delete paths

MC-Meesh:
Completes #437 (the GC-scoping half landed in #457): the six workload adoption/update/delete paths in the App reconciler now check Mortise ownership before touching anything.

## What

- New `conflictIfUnmanaged` guard + typed `resourceConflictError`: a pre-existing resource carrying one of an App's reserved names without the `app.kubernetes.io/managed-by: mortise` label is refused — never adopted, updated, or deleted. Checked **twice** per path (after the initial read AND inside the conflict-retry closure), mirroring the ConfigMap/credentials-Secret paths that already did this correctly, so a foreign resource created between read and write is still refused.
- Guarded paths: Deployment (guard placed **before** the selector-mismatch delete — previously a foreign Deployment's non-matching selector got it deleted on sight), CronJob, Service, ServiceAccount (previously its `imagePullSecrets` could be silently swapped, redirecting a foreign workload's image pulls), Ingress (wholesale spec replace), PVC (resize + relabel).
- Conflicts surface on the App: `envResourceError` maps the typed error to `Failed` + a `ResourceConflict` condition with an actionable message (rename or delete the conflicting resource). The condition clears on the next clean pass, same pattern as `DomainCollision` and `RBACForbidden`.

## Why hard-fail instead of skip-and-log

A silent skip leaves the App half-deployed with no signal. The condition tells the user exactly which resource is in the way; deleting or renaming it un-wedges the App on the next reconcile (covered by the recovery test).

## Tests

Seven envtest specs: one per guarded path (foreign resource survives untouched, byte-for-byte assertions on the fields the old code would have destroyed, App gets `Failed` + `ResourceConflict`) plus a recovery spec (delete the foreign resource → condition clears, App proceeds, resource is recreated managed).

## Gates

`make test`, `go vet`, staticcheck v0.7.0 (CI pin), `make test-charts` — green. `make test-integration`: 54/57 pass; the 3 timeouts (`TestPreviewInheritsStagingBindings`, `TestFromBindingProjectsNamedKey`, `TestObserverTraffic`) are root-caused to a pre-existing k5p residual on main — `reconcileResolvedEnvSecrets` (git-preflight path) bypasses the #464 classifier, the 12th env-write call site (tracked as mo-tpe with operator-log evidence). The guards in this PR never fired in those runs: zero `not managed by Mortise` hits across 1340 operator log lines. This branch neither touches nor could fix that path.

Fixes mo-p2l. Ref #457, #464 (composes with the RBAC-race classifier in `envResourceError`).

Note for a follow-up (not this PR): recovery latency after the user deletes the foreign resource rides the error-backoff requeue — foreign resources carry no Mortise labels, so their deletion doesn't enqueue the App, and worst-case backoff is ~16 min. A `RequeueAfter`-style poll like `domainCollisionRequeueInterval` would bound it at a minute; noted on the bead.

### DIFF
--- a/internal/controller/app_controller.go
+++ b/internal/controller/app_controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -100,6 +101,14 @@ const rbacPropagationRequeue = 2 * time.Second
 // from writing into an env namespace old enough that RBAC propagation cannot
 // explain it.
 const appRBACForbiddenCondition = "RBACForbidden"
+
+// appResourceConflictCondition is set when a workload reconcile finds a
+// pre-existing resource carrying one of this App's reserved names without the
+// Mortise managed-by label. Per CLAUDE.md "Mortise owns only what it creates"
+// the reconciler refuses to adopt, update, or delete such a resource; the
+// condition makes the conflict visible on the App instead of only in operator
+// logs.
+const appResourceConflictCondition = "ResourceConflict"
 
 // AppReconciler reconciles a App object
 type AppReconciler struct {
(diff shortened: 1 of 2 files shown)


MC-Meesh:
**Cross-review verdict: APPROVE with one should-fix and one note** (posted as comment — same-account PRs cannot be formally approved).

Reviewed against the mo-p2l packet with the branch checked out locally: full controller suite green (envtest, 36s), staticcheck clean.

**What's right.** The guard shape is exactly correct: `conflictIfUnmanaged` at both the initial read AND inside every conflict-retry closure closes the create-race window (a foreign resource appearing between read and retry is still refused); the terminal `resourceConflictError` → `envResourceError` → Failed + `ResourceConflict` condition composes cleanly with the #464 classifier ordering (conflict checked before the Forbidden branch, correct — a conflict is never transient), and the clear-on-recovery block mirrors the RBACForbidden pattern. The Deployment guard placement before the selector-mismatch delete kills the worst hazard (delete-on-sight of a foreign Deployment). The seven specs cover all six paths plus recovery.

**Should-fix: `reconcileExternalNameService` is an unguarded seventh path** (app_controller.go ~:4406-4428). It has the same shape this PR fixes elsewhere, with the same worst case: a foreign Service under the App's reserved name gets **deleted** on type mismatch (`existing.Spec.Type != ExternalName` → Delete + Create) or blind-updated otherwise. External-source apps are exactly the kind to collide with hand-made Services. Same two-line guard + one spec.

**Note (no action needed):** the guard hard-fails any Mortise-created resource that predates the managed-by label — but label stamping is contained in v0.1.0 (verified via `git tag --contains` on the introducing commit), so no real install has unlabeled Mortise workloads. Worth one CHANGELOG line under the existing #457 label-contract entry, since this extends that contract from GC to adopt/update/delete.

MC-Meesh:
**Re-review: the should-fix is correctly resolved — APPROVE stands** (comment-approval, same-account). The ExternalName guard sits before the type-change delete exactly like the Deployment one; the new spec pins the worst case (foreign ClusterIP Service survives untouched, App reports ResourceConflict). Your single-post-Get-check reasoning is sound: no retry closure exists on that path, and a lost Create race surfaces as an error whose next reconcile re-runs the guard. Verified locally at 509d3bf: full controller envtest suite green.

Issue #214, opened 2026-05-07
feat: PVC storage metrics collection and display


JamesSprow:
## Summary

Investigate and implement PVC storage usage metrics — actual bytes used vs capacity, not just requested size. This is currently a gap in the observability stack: we know what PVCs exist and their requested capacity (from PVC spec), but not actual utilization.

## Motivation

- The dashboard page (#210) needs storage metrics alongside CPU/memory/traffic. Storage is deferred from dashboard V1 pending this work.
- Users with backing services (Postgres, Redis) running persistent volumes need to know when they're approaching capacity before it becomes an outage.
- Capacity planning requires actual usage data, not just provisioned amounts.

## Investigation Areas

1. **kubelet stats summary API** (`/stats/summary`) — exposes per-volume `usedBytes`, `capacityBytes`, `availableBytes`. Requires node-level access. How to aggregate across nodes efficiently?
2. **CSI metrics** — some CSI drivers expose volume metrics via Prometheus. Could the observer scrape these if available?
3. **metrics-server** — only covers CPU/memory for pods, not volumes. Not a path here.
4. **PVC status conditions** — some storage providers populate `.status.capacity` but not usage.

## Design Questions

- Should this live in the observer (new storage collector alongside metrics/traffic/logs collectors)?
- What RBAC is needed? kubelet stats API requires node proxy access.
- How to handle storage classes that don't expose usage (graceful degradation)?
- Retention policy — storage usage changes slowly, 5s polling is overkill. What interval?

## Relationship to Dashboard (#210)

Dashboard V1 defers the storage section. Once this is implemented, the dashboard should show:
- Per-app PVC usage bars (used / capacity)
- Per-project storage aggregation
- Cluster-wide storage totals (admin view)

The dashboard API response has a placeholder for storage with `"available": bool` — this issue provides the data source.

## Out of Scope

- Storage provisioning/resizing through Mortise (we don't manage PVCs beyond what apps declare)
- Cost estimation for storage

PR #467, opened 2026-08-10
feat(observer): gap-visible pipeline, restart resilience, retention, PVC usage (obs-v2 O1)

MC-Meesh:
## Summary

Obs-v2 foundation (tracker bead mo-5i4): the observer's per-app series become trustworthy. The audit behind the 'shotty tracking' verdict produced a **twelve-gap written list** — the deliverable itself, in `docs/observer-reliability.md` — and this PR fixes every non-by-design entry.

**The reliability contract:**

1. **Gaps are visible.** Every collector cycle writes a heartbeat (`collector_ticks`); a window with a successful heartbeat but no series rows was *observed empty*, one without a heartbeat was *not observed*. `/v1/metrics` and `/v1/pvc` gain a `coverage` array (`[bucket, 0|1]`) — the contract the O4 UI renders as gaps, never interpolating. Collector failures (namespace list, per-ns metrics, inserts) record failed ticks with causes instead of log-and-continue silence; zero-entry cycles are recorded as valid observed-empty data.
2. **Restarts resume.** Log tailers resume from a per-pod cursor (`SinceTime` from the last stored line, truncated down one second: bounded ≤1s re-read, never a skip) — previously every tailer restart re-ingested up to 100 duplicate lines and lost anything beyond 100 written during downtime. Traffic flushes all open buckets at shutdown (previously up to a full bucket died per restart) and failed insert batches retry next flush instead of being dropped.
3. **Storage is bounded.** Policy decided and implemented: raw metrics for 24h → idempotent 5-minute downsampling → age cutoff; heartbeats/PVC follow the metrics window. Long-horizon storage stays explicitly out of scope (that's O2's Prometheus surface).
4. **Self-health.** `/v1/health/collectors`: per-collector last tick/success/error + gauges (`logTailers`, `logTailersSkipped` — the silent maxPods cap is now visible — and cumulative `logLinesDropped`). Dashboards can say 'metrics stale since X' instead of flatlining.

**PVC usage (#214):** new collector samples each node's kubelet Summary API — the only interface reporting actual per-volume filesystem usage — keyed `(namespace, app, env, pvc)`, served at `/v1/pvc` under the same coverage contract. Degrades to failed-heartbeats + absent metrics where clusters block `nodes/proxy`. **RBAC**: read-only `nodes` get/list + `nodes/proxy` get, justified inline in `observer-rbac.yaml` — which lives in the umbrella chart, so no collision with mo-e1z's mortise-core file.

Additive API only: the operator's raw-passthrough proxy carries the new `coverage` field with no operator-side change; `/v1/pvc` and the health endpoint await O2/O4 surfacing.

## Verification

- Observer suite green including **ten new reliability tests**: failure-injected ticks, observed-empty ticks, coverage bucketing (incl. failed-tick ≠ coverage), cursor resume bounds, traffic retain-on-failure + drain-after-recovery, downsample collapse + idempotence, trim of new tables, PVC degrade + labeled-series success paths.
- `make test-charts` and ci-local fast gates green.
- Full-gate integration: two runs hit the **load-dependent ready-ceiling flake now filed as bead mo-j6k** (zero forbidden errors — confirmed NOT the mo-k5p/mo-tpe signature; isolated `TestFullStackDeploy` repro passes in 118s with the same doubled-RS pattern, proving the mechanism is pre-existing mid-startup rollout × RWO PVC handoff under `-parallel 25`, unrelated to this branch's observer-only diff). A third run's collapse was an in-container k3s restart (no OOM, RestartCount 0). Full forensics on the bead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- **Observer reliability + PVC usage** (obs-v2 O1, #214): two new observer
+  endpoints — `GET /v1/pvc` (per-PVC capacity/usage series, collected from
+  the kubelet Summary API) and `GET /v1/health/collectors` (per-collector
+  last tick/success/error plus tailer and dropped-line gauges). `/v1/metrics`
+  and `/v1/pvc` responses gain a `coverage` array marking observed vs
+  unobserved windows so UIs render gaps instead of interpolating. The
+  observer ClusterRole widens by read-only `nodes` get/list and
+  `nodes/proxy` get for kubelet stats (justified inline in the chart).
+  Retention change: raw metrics downsample to 5-minute averages after 24
+  hours (was raw for the full 72h window); log tailers now resume from a
+  stored cursor across restarts instead of re-reading (and duplicating)
+  the last 100 lines. Full audit: `docs/observer-reliability.md`.
 - **Published-chart verification in the release pipeline** (#446, #379
   residual): after publishing to gh-pages, `release.yml` pulls the chart
   back through the public repo URL, asserts it is byte-identical to what
--- a/SPEC.md
+++ b/SPEC.md
@@ -1126,18 +1126,25 @@ Mortise UI. Default retention:
 | Data type | Retention |
 |---|---|
 | Logs | 48 hours |
-| Metrics | 72 hours |
+| Metrics | 72 hours (raw for 24h, then 5-minute averages) |
 | Traffic | 48 hours |
+| PVC usage | 72 hours (follows metrics) |
 
 Observer HTTP endpoints (served at the adapter base URL):
 
 | Endpoint | Description |
 |---|---|
 | `GET /v1/logs` | Historical log lines (time range query) |
-| `GET /v1/metrics` | Historical metrics |
+| `GET /v1/metrics` | Historical metrics (includes `coverage` gap markers) |
 | `GET /v1/metrics/live` | Live metrics stream (SSE) |
 | `GET /v1/traffic` | Historical traffic data |
 | `GET /v1/traffic/live` | Live traffic stream (SSE) |
+| `GET /v1/pvc` | Historical per-PVC capacity/usage (includes `coverage`) |
+| `GET /v1/health/collectors` | Observer self-health: per-collector last tick/success/error |
+
+`coverage` is the gap-visibility contract: `[bucketTs, 1]` marks a window the
+collector observed (even if it found nothing), `[bucketTs, 0]` one it did not.
+Consumers render 0-buckets as gaps and never interpolate across them.
 
 The observer is wired in via `PlatformConfig.spec.observability`:
 `metricsAdapterEndpoint`, `logsAdapterEndpoint`, `trafficAdapterEndpoint`.
--- a/charts/mortise/templates/observer-rbac.yaml
+++ b/charts/mortise/templates/observer-rbac.yaml
@@ -28,6 +28,16 @@ rules:
   - apiGroups: [metrics.k8s.io]
     resources: [pods]
     verbs: [get, list]
+  # PVC usage collection (#214): the kubelet Summary API is the only
+  # interface reporting actual per-volume filesystem usage, and it is
+  # reachable only via nodes/proxy (GET stats/summary). Read-only; the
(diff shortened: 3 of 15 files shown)


MC-Meesh (commented):
**Verdict: APPROVE with one convention should-fix (CHANGELOG)** (posted as comment — same-account PRs cannot be formally approved).

Cross-reviewed against the mo-5i4/O1 packet. Every packet item is delivered, and the deliverable I most wanted — the written gap list with teeth — exists as `docs/observer-reliability.md` G1–G12, each mapped to code and fix. The core design decision is exactly right: `collector_ticks` heartbeats make *observed-empty* and *not-observed* first-class distinct states, the `coverage` array exports that to consumers as the no-interpolation contract, and the tests pin the semantics from both directions (observed-empty is a successful 0-item tick; a *failed* tick is explicitly not coverage). Restart downtime renders as 0-coverage buckets for free since no ticks were written — the right behavior falling out of the model rather than being bolted on.

**Verified against the packet, item by item:** gap-visible model (G1–G4, tested); restart resilience — the log cursor resume is thoughtfully bounded (truncate-down to the second: ≤1s re-read, never a skip, trade documented in place) and traffic gets both shutdown `flushAll` and the pending-batch retry so a transient SQLite error delays rather than loses (G5–G7, both tested including the recover-and-drain path); retention with idempotent in-place downsampling and the policy table in the doc (G8, tested for collapse, averages, and no-op second pass); PVC collection via kubelet Summary with the RBAC justified by comment in the chart — and correctly placed in the umbrella's own `observer-rbac.yaml`, so no mo-e1z collision and no subchart-tgz repack applies; graceful degradation is a failed heartbeat plus absent metrics, tested; self-health endpoint with the tailer/skipped/dropped gauges closes G9–G11, making the `maxPods` silent un-tailing visible at last.

**Should-fix before merge: CHANGELOG entry.** Two new public endpoints (`/v1/pvc`, `/v1/health/collectors`), a `coverage` field on `/v1/metrics`, a ClusterRole widening (`nodes`, `nodes/proxy`), and a retention/downsampling behavior change are all user-facing; REVIEW.md §14 wants them under `[Unreleased]`. One entry covering the reliability contract + PVC metrics does it. The SPEC §5.11a adapter-contract endpoint list should pick up the two new GETs too — fine as a ride-along line here or in O2's SPEC work, but say which.

**Non-blocking observations:**
1. `LastLogTimestamp` uses SQL `MAX(ts)` on RFC3339Nano *text*. Lexicographic max mis-orders same-second timestamps with differing fractional-digit counts (`…05Z` sorts above `…05.5Z`). Today this is harmless — the cursor is truncated down to the second, so the worst case stays inside the documented ≤1s re-read — but it's a latent trap for any future caller using the value untruncated. A one-line comment on the function would immunize it.
2. `pending` in the traffic collector grows unboundedly if the store stays broken. Acceptable for the transient-error design (persistent store failure is loudly visible in health), but a cap-with-counter would make the memory bound explicit.
3. Coverage is exported on `/v1/metrics` and `/v1/pvc` but not `/v1/traffic`, whose heartbeats exist. Traffic charts interpolate as much as metrics charts do — worth a small follow-up so O4 can honor the gap contract on all three series types (logs don't interpolate; N/A).
4. Two micro-nits in `downsampleMetrics`: the `COUNT(1) … LIMIT 1` pending-check counts every row where `EXISTS` semantics would do, and a poll interval that divides 300s exactly could leave boundary-aligned raw rows uncollapsed (pathological config only — 5s/60s defaults are fine).
5. G6's fix covers graceful shutdown; a SIGKILL/OOM still loses the open traffic bucket (≤ bucket size). That's inherent to memory aggregation and fine — but it belongs in the doc next to G12 as a stated bound rather than an unstated one.

The audit-doc-plus-contract-tests shape of this PR is what the epic's "reliable over pretty" verdict needed as its foundation. With the CHANGELOG entry added, ready for the train.

MC-Meesh:
Review items addressed in 071ab9f: CHANGELOG entry covers both new endpoints, the `coverage` field, the read-only nodes/nodes-proxy RBAC widening, and the retention/downsampling change. **SPEC §5.11a judgment call: updated in this PR** — the endpoint table is a statement of what ships and should never lag reality; O2 can restructure a true table as easily as a stale one (stated in the commit message too). crew2's traffic-coverage observation is filed as bead mo-vkl (extend the gap contract to /v1/traffic, decide the /v1/logs shape) — flagged as an O4 prerequisite.

PR #468, opened 2026-08-10
fix(controller): classify the git-preflight env-secret path and bound the missing-namespace phase

MC-Meesh:
Kills the last known k5p residual and bounds the classifier's missing-namespace phase. Fixes mo-tpe. Fixes mo-4po. Ref mo-k5p.

## The 12th call site (mo-tpe)

`reconcileResolvedEnvSecrets` runs on the git-source **preflight** path — bindings and shared vars must materialize even when `prepareGitSource` says don't-proceed — which means it writes into freshly bootstrapped env namespaces and races the same RBAC propagation as the main env loop. Its errors returned plain, feeding the workqueue's exponential rate limiter: the mo-k5p compounding, alive on trees that already carried the #464 classifier (field evidence: 31 forbidden / 3 timeouts on the mo-p2l gate run, archived under `audit-2026-08-09/mo-p2l-gate/`). Its errors now route through `envResourceError` like the other eleven sites; the helper returns `(ctrl.Result, error)` and the caller preserves preflight-error-wins semantics. The existing `reconcile env secret for env X` message shape is preserved.

## Bounding phase C (mo-4po)

The missing-namespace arm fast-requeued forever: a namespace that never exists has no age to bound against, so a genuinely broken bootstrap became a silent 0.5 Hz poll with the App never reaching Failed. Now the first missing-namespace hit stamps a `NamespacePending` condition (`LastTransitionTime` from the injected clock — `meta.SetStatusCondition` otherwise stamps wall-clock, which a fake-clock test catches as a negative duration); repeated hits keep the original timestamp, which *is* the escalation timer; past `namespaceCreationBudget` (2 min — deliberately a separate constant from `rbacPropagationWindow`: it times namespace creation by the Project controller, not authorizer propagation) the App goes `Failed` with `NamespaceCreationTimeout` and an actionable message. The condition clears on the next clean pass alongside `RBACForbidden`. The shared `forbiddenFastRequeue` keeps its unbounded NotFound arm with a comment proving its remaining callers cannot loop on it (the PE controller checks `previewNamespaceNotReadyError` first; the Project controller passes a namespace it ensured in the same reconcile).

## Tests

Five new unit tests (phase-C condition set, within-budget requeue with surviving transition time, past-budget escalation, preflight young-ns fast-requeue, preflight old-ns `Failed` + `RBACForbidden`, non-forbidden passthrough pinning the message shape) plus an envtest spec for clean-pass condition clearing.

## Gates (full `make ci-local`)

unit+envtest 68s, helm lint, vet+staticcheck, ui build, **ui-e2e 137s** — all pass. Integration: 55/57, with **the fix's own signature provably dead**: all 21 forbidden events in the operator log are INFO fast-requeues (including the new preflight classification firing on a preview app), zero error-path forbidden, zero rate-limiter feeding. The two failures are pre-existing filed flakes — the known preview-webhook flake (mo-24b) and the zero-forbidden load flake crew1 just root-caused (mid-startup env-hash rollout × RWO PVC handoff under `-parallel 25`) — and **both pass in an isolated rerun (81s) on the same tree and cluster**. Gate log + forbidden-line extract archived under `audit-2026-08-09/tpe-gate/`.

Merge note: touches `envResourceError` and the clean-pass clearing block — the same regions PR #466 extends (`ResourceConflict` branch). Whichever merges second rebases trivially; the hunks are adjacent, not overlapping.

### DIFF
--- a/internal/controller/app_controller.go
+++ b/internal/controller/app_controller.go
@@ -101,6 +101,22 @@ const rbacPropagationRequeue = 2 * time.Second
 // explain it.
 const appRBACForbiddenCondition = "RBACForbidden"
 
+// appNamespacePendingCondition is set while an env write is denied because the
+// env namespace does not exist yet (the authorizer rejects before existence is
+// checked, so a write racing namespace bootstrap surfaces as Forbidden). Its
+// LastTransitionTime is the bound: a namespace has no age of its own to bound
+// against, so the condition's first-set time stands in for it. Past
+// namespaceCreationBudget the App goes Failed — a namespace that never
+// appears means the Project controller cannot create it, which is a genuine
+// misconfiguration, not a race.
+const appNamespacePendingCondition = "NamespacePending"
+
+// namespaceCreationBudget bounds how long an App fast-requeues on
+// Forbidden-in-a-missing-namespace before escalating. Kept separate from
+// rbacPropagationWindow (same default) because it times a different thing:
+// namespace creation by the Project controller, not authorizer propagation.
+const namespaceCreationBudget = 2 * time.Minute
+
 // AppReconciler reconciles a App object
 type AppReconciler struct {
 	client.Client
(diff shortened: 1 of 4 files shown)


MC-Meesh:
**Cross-review verdict: APPROVE** (comment-approval, same-account). Reviewed with the branch checked out: full controller envtest suite green (35s), staticcheck clean.

**mo-tpe (12th call site)**: exactly right. `reconcileResolvedEnvSecrets` returns `(ctrl.Result, error)` and routes per-env failures through `envResourceError`, closing the last bypass of the classifier. The preflight call-site precedence is the correct judgment: when `prepareGitSource` itself errored, that error still wins (a genuine failure must not be hidden behind an errorless requeue) and the env-secret outcome is logged rather than lost; when preflight merely said don't-proceed, the fast-requeue propagates properly. As the author of the other eleven wraps I'll own the miss: this site's multi-env signature made it awkward for the single-ns helper and I deferred it — your Result-returning shape is the right generalization.

**mo-4po (phase-C bound)**: the condition-as-timer design is verified-correct — `meta.SetStatusCondition` preserves `LastTransitionTime` while Status stays True, so the first-hit timestamp is the escalation clock, stamped from the injected clock for testability; the clearing block resets it on any clean pass so each race episode times independently. Escalation past the budget to Failed/`NamespaceCreationTimeout` with a plain error (feeding the limiter) is right — a namespace that never appears is a misconfiguration, and the message points at the actual culprit (Project controller status + RBAC). The `forbidden.go` comment documenting why the shared helper's missing-ns arm stays unbounded for the other two callers checks out against both call sites (PE's `previewNamespaceNotReadyError` pre-check; Project passing a namespace it ensured in the same reconcile).

**Tests**: the six new specs cover both halves precisely (pending-set, budget-escalation, and the young/old/non-forbidden trio for the new site) plus the envtest clean-pass clearing spec. Your gate evidence (21/21 forbidden classified as INFO fast-requeues; the two integration reds being mo-24b + mo-j6k, both passing isolated) matches my own runs' characterization.

No findings. Ship it.

PR #476, opened 2026-08-11
feat(observability): Prometheus exposition on observer and operator, Grafana recipe (obs-v2 O2)

MC-Meesh:
## Summary

Obs-v2 O2 (tracker bead mo-skg): everything the built-in dashboard shows becomes scrapeable, and power users get Grafana with zero Mortise changes.

## The metric set (public API — review this list)

**Observer `GET /metrics`** (hand-rolled text exposition, zero new deps; `{project, app, environment}` labels throughout):

| Metric | Type |
|---|---|
| `mortise_app_cpu_cores` / `mortise_app_memory_bytes` | gauge (+ `pod`) |
| `mortise_app_pod_restarts_total` | counter (+ `pod`) |
| `mortise_app_pvc_capacity_bytes` / `mortise_app_pvc_used_bytes` | gauge (+ `pvc`) |
| `mortise_app_http_requests_total` | counter |
| `mortise_app_http_responses_total` | counter (+ `class`) |
| `mortise_app_http_request_bytes_total` / `_response_bytes_total` | counter |
| `mortise_app_http_latency_seconds` | gauge, latest bucket (+ `quantile`) |
| `mortise_observer_collector_up` / `_last_success_timestamp_seconds` | gauge (+ `collector`) |
| `mortise_observer_log_tailers` / `_skipped` / `_lines_dropped_total` | gauge / counter |

**Operator `/metrics`** (controller-runtime endpoint — no observer duplication):

| Metric | Type |
|---|---|
| `mortise_builds_total` | counter (`project, app, result`) |
| `mortise_build_duration_seconds` | histogram (same labels) |
| `mortise_app_status_phase` | gauge, 1 per app (`project, app, phase`) |

Semantics: counters reset on restart (prom-normal, advanced only on durable traffic flush so pending-retries can't double-count); stale series age out in ~5min; exposition is deterministically ordered and format-tested (HELP/TYPE, escaping, label order).

## Audit finding (bead item 2)

The operator's metrics endpoint existed in code but was **disabled by default** (`--metrics-bind-address=0`) and unexposed in the chart — nothing could ever have scraped it. The chart now binds it plain-HTTP (cluster-internal read-only telemetry, same posture as the observer), exposes the port + scrape annotations, and adds the build/phase metrics above (build metrics instrumented at the BuildRun terminal transition; the phase gauge is a cache-backed collector costing no apiserver calls per scrape).

## Scrape discovery

`prometheus.io/*` annotations on both Services (default **on**); ServiceMonitors for kube-prometheus-stack behind `observer.prometheus.serviceMonitor.enabled` (default **off** — the CRD doesn't exist without prometheus-operator), with `honorLabels: true`.

## Grafana

`docs/recipes/grafana.md` (kps + plain scrape configs, query cookbook, semantics notes) and an import-ready `docs/recipes/mortise-dashboard.json` (cluster overview + per-app drilldown, `project`/`app`/`environment` variables).

## Verification

- **Live against a real kube-prometheus-stack on k3d**: ServiceMonitor discovered, both targets `up`, PromQL returned labeled series from a real running app (`mortise_app_cpu_cores{project="obs-demo",...}`, `mortise_app_status_phase{...,phase="CrashLooping"}` — the latter proving live state), dashboard imported through the Grafana API with a panel query returning data. The live run surfaced and fixed two integration bugs specs alone would have missed: kps's honor-labels default renaming the app `pod` label to `exported_pod`, and the operator's secure-serving default speaking HTTPS at an HTTP ServiceMonitor.
- Observer suite green including 6 new exposition tests; `make test-charts` green (both ServiceMonitor toggles templated); drift check green (mortise-core repacked).
- Full `make ci-local`: 5/6; integration red was exactly mo-24b + the mo-j6k load shape with **zero error-level forbidden** (22 denials, all classified fast-requeues — incidental live proof of #464/#468 in the wild). No overlap with this diff.

SPEC §5.11b records the names as a compatibility contract. Fixes the mo-skg tracker bead.


MC-Meesh (commented):
**Verdict: APPROVE with two should-fixes (a Helm false-trap and the CHANGELOG)** (posted as comment — same-account PRs cannot be formally approved).

Cross-reviewed against the mo-skg/O2 packet. The shape is exactly the pluggable-depth design the epic called for, and several decisions deserve naming: the hand-rolled exposition writer with its stated rationale (the observer binary staying free of the operator's dependency tree is worth more than a client library for what are all point-in-time gauges and simple counters); `honorLabels: true` on both ServiceMonitors with the comment explaining the `exported_pod` renaming trap — a catch most first passes miss; traffic counters advancing **only on durable flush** so the pending-retry path can't double-count (composing correctly with O1's retention machinery); the SPEC §5.11b table making the metric set an explicit compatibility contract, which is what my packet's "treat naming as public API" item actually wanted; and the vendored `mortise-core-0.0.0-dev.tgz` repacked alongside the subchart template changes — the trap remembered without being reminded. Verified off-diff: the new operator metrics port (8080) does not collide with `api.port` (8090), and both `--metrics-bind-address` and `--metrics-secure` exist in `cmd/main.go` with secure defaulting *true* — so the chart's `--metrics-secure=false` is required, not decorative.

**Should-fix 1 — Helm's `default` false-trap in `operator-servicemonitor.yaml`.** The guard is `index .Values "mortise-core" "metrics" "enabled" | default true`. Sprig's `default` treats `false` as empty, so a user who *explicitly disables* operator metrics (`mortise-core.metrics.enabled=false`) still gets `true` out of that pipeline — and with `serviceMonitor.enabled=true` the chart renders a ServiceMonitor pointing at a port the Deployment no longer exposes (the deployment template gates correctly, so the two templates disagree). Narrow, but it's precisely the silent-values-mismatch class #446 taught us to respect. One-line fix: compare explicitly (`ne ... false`) or `hasKey`-guard instead of `default`.

**Should-fix 2 — CHANGELOG entry** (same REVIEW.md §14 miss as #467's first revision): two new public `/metrics` endpoints, a chart-values surface (`metrics.*`, `observer.prometheus.*`), new ServiceMonitor templates, and a SPEC public-API section are all user-facing. One entry covers it.

**Non-blocking observations:** (1) In the metrics collector, a failed per-namespace pod list zeroes `restarts` for every pod that cycle — `mortise_app_pod_restarts_total` then dips to 0 and back, which `rate()` reads as a counter reset plus a spurious spike. Omitting the sample (keep the previous value in `promState`) on list failure would be more honest than zeroing. (2) `mortise_app_http_latency_seconds` as quantile-labeled gauges of the latest bucket is a defensible lightweight choice, and the Grafana doc's "graph them raw, don't average" warning is exactly the right disclosure — keep that doc section when anyone touches this. (3) For the record: `internal/controller/metrics.go` imports `prometheus/client_golang`, which brushes the letter of the no-third-party-SDKs-in-controllers red flag. It's the sanctioned mechanism — controller-runtime's own `metrics.Registry` *is* a `*prometheus.Registry`, so custom operator metrics cannot exist without these types, and an indirection layer here would be abstraction theater. But the rule should say that rather than be silently excepted: suggest a one-line carve-out in CLAUDE.md/REVIEW.md ("k8s-ecosystem libraries that controller-runtime itself exposes in its API are permitted") in this PR or a docs follow-up, so the next reviewer doesn't have to re-derive the judgment.

The live verification (kube-prometheus-stack scrape + dashboard import on a real cluster) was the packet's highest-cost item and it's what makes the recipe trustworthy. With the two should-fixes, train-ready.

MC-Meesh:
Review items addressed in e887555. (1) **The `default true` bug is real and fixed properly**: parent templates can't see subchart defaults (which is why the sprig `default` was there), so the default is now mirrored in the umbrella values (`mortise-core.metrics.enabled: true`) and the condition reads the real value — verified both directions by template: explicit `false` renders zero operator ServiceMonitors, default renders one. (2) CHANGELOG entry added covering both endpoints, the operator exposure, scrape wiring, and the Grafana recipe. (3) crew2's governance ask codified in REVIEW.md §12: `client_golang` is operator-only (controller-runtime hosts the registry); the observer's exposition stays hand-rolled, with extension guidance for both sides. `make test-charts` green.

PR #477, opened 2026-08-11
feat(activity): measured cap raise, cursor pagination, time pruning (obs-v2 O3)

### LINKED ISSUE #165: feat: activity store scalability — replace ConfigMap ring buffer
## Problem

The activity store (`internal/activity/configmap_store.go`) uses a ConfigMap-based ring buffer capped at **500 events per project**. Once the buffer is full, the oldest events are silently dropped. High-velocity projects (frequent deploys, many apps, active preview environments) can blow through 500 events in hours, losing audit history.

The SPEC (§10 item 6) acknowledges this: "A project that generates >500 events/day will see entries fall off the rail faster than users can read them."

## Current implementation

- One ConfigMap per project in the control namespace (`pj-{project}`)
- Events stored as JSON array in the ConfigMap's `data.events` key
- Ring buffer trims oldest entries when count exceeds 500
- ConfigMap size limit is ~1MB (Kubernetes hard limit), which also caps total storage
- No pagination in the API — the UI loads all 500 events at once

## Impact

- **Audit trail is unreliable.** Users can't scroll back to see what happened yesterday if >500 events occurred since.
- **Compliance gap.** Any team needing deploy audit trails can't rely on the activity rail.
- **No export.** Events that fall off the ring buffer are gone — they're not in logs or anywhere else.

### Option A: SQLite-backed activity store (like the observer)

The observer already uses SQLite for metrics/logs/traffic storage. The activity store could use the same pattern:
- SQLite database in a PVC (or the same PVC as the observer)
- Proper time-based retention (e.g., 30 days) instead of count-based
- Pagination support in the API
- Much higher capacity (SQLite can handle millions of rows)

### Option B: Structured log emission + external aggregation

- Emit activity events as structured JSON logs to stdout (the operator already logs to stdout)
- Users pipe to Loki/Elasticsearch/CloudWatch for long-term storage
- Keep the ConfigMap ring buffer for the "quick glance" UI rail (short-term cache)
- Document the log format so users can build dashboards

### Option C: Increase ConfigMap cap + add pagination

- Raise the limit from 500 to ~2000 (still within 1MB ConfigMap limit depending on event size)
- Add pagination to the activity API (`?page=1&limit=50`)
- Cheapest fix but doesn't solve the fundamental problem

### Recommendation

**Option A for the primary store** (SQLite with time-based retention gives real audit history) **+ Option B as a complement** (structured logs let users integrate with existing observability stacks). Option C is a quick band-aid if the others are deferred.

## Acceptance criteria

- [ ] Activity events are retained for a configurable duration (e.g., 30 days), not a fixed count
- [ ] Activity API supports pagination
- [ ] High-velocity projects don't lose recent history
- [ ] Migration path from ConfigMap store to new store (or graceful fallback)

MC-Meesh:
Obs-v2 O3: the ConfigMap activity store keeps its architecture and loses its arbitrary limits. Fixes #165 (Option C). Fixes mo-1mm.

## Measured cap raise (the math)

`Cap` goes 500 → 2000, and the number is derived, not guessed: `TestCapFitsEtcdBudget` marshals 2000 representative events and asserts the payload fits the budget. Measured typical entry: ~260 B; 2000 of them ≈ 500 KiB. The real guarantee is the new **`MaxBytes` budget (768 KiB)** — leaving ≥256 KiB of the 1 MiB etcd object limit for ConfigMap metadata and managedFields — enforced on every append by oldest-first byte-trim, so pathological entry sizes degrade capacity but can never block a write (tested with 8 KiB messages). If `Event` grows and typical size drifts, the math test fails and the constants get re-derived instead of silently lying.

## Cursor pagination

Every event gets a per-project monotonic `seq` at append (with a one-time backfill for pre-existing zero-seq entries, applied in-memory on reads so `List` never mutates). `ListPage` returns seq-anchored pages that are **stable across concurrent appends and trims** — the pagination test appends mid-pagination and asserts page 2 continues exactly where page 1 ended, no duplicates, no skips. The API stays backward-compatible: the response body remains a bare array; the next-page cursor travels in the `X-Next-Cursor` response header (absent on the last page), `?cursor=` feeds it back, invalid cursors are a 400. Page size is capped at 500 — a full buffer is ~500 KiB of JSON, which is what pagination is for.

The platform-wide endpoint stays limit-only on purpose: per-project cursors don't compose across a merged stream, and the platform dashboard (obs-v2 O5) gets its own aggregate endpoint. Rationale is in the handler comment.

## Time pruning

Entries older than 30 days drop on append — which already rewrites the whole ConfigMap, so pruning is free. Retention never drops the event being appended (tested). The stdout audit stream remains the unbounded system of record.

## UI

The rail loads pages and gains a **Load more** button (exact-name selector discipline). New `activity-rail.spec.ts` covers open, shows recorded activity, close as baseline e2e; the Load-more control only renders past one page (100+ events), which real-API seeding can't reasonably produce in e2e — its logic is pinned at the store and handler layers instead, stated in the spec's header comment.

## Explicitly future

The SQLite/observer-backed store (#165 Option A) is **not** this PR and stays deferred until obs-v2 needs it: the `Store` interface deliberately stays narrow (`Append`/`List`/`ListPage`) so that backend is a drop-in swap.

## Gates

Full ci-local, run in two halves coordinated around crew1's cluster usage: unit+envtest 79s, helm lint, vet+staticcheck, ui build — green; `make test-integration` fully green (523s, zero failures); ui-e2e 161/161 (42.8s) including the new rail spec. Swagger regenerated.

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -18,6 +18,13 @@ definitions:
         type: string
       resource:
         type: string
+      seq:
+        description: |-
+          Seq is a per-project monotonically increasing sequence number assigned
+          at append time. It makes pagination cursors stable when timestamps
+          collide. Entries written before pagination existed carry Seq 0 until
+          the one-time backfill in appendOnce assigns them positions.
+        type: integer
       ts:
         type: string
     type: object
@@ -3132,22 +3139,32 @@ paths:
       - projects
   /projects/{project}/activity:
     get:
-      description: Returns recent activity events for a project, newest first
+      description: Returns recent activity events for a project, newest first. The
+        X-Next-Cursor response header carries the cursor for the next page; pass it
+        back via the cursor query parameter. No header means no further pages.
       parameters:
       - description: Project name
         in: path
         name: project
         required: true
         type: string
-      - description: 'Max number of events (default: 100)'
+      - description: 'Max number of events per page (default: 100, max: 500)'
         in: query
         name: limit
         type: integer
+      - description: Opaque cursor from a previous page's X-Next-Cursor header
+        in: query
+        name: cursor
+        type: string
       produces:
       - application/json
       responses:
         "200":
           description: OK
+          headers:
+            X-Next-Cursor:
+              description: Cursor for the next page; absent on the last page
+              type: string
           schema:
             items:
               $ref: '#/definitions/activity.Event'
--- a/internal/activity/configmap_store.go
+++ b/internal/activity/configmap_store.go
@@ -5,10 +5,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"strconv"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -17,7 +18,28 @@ import (
 )
 
 // Cap is the ring-buffer size: the maximum number of events kept per project.
-const Cap = 500
+//
+// The bound is the etcd object-size limit (1 MiB per ConfigMap, including
+// metadata and managedFields). Measured marshaled entry sizes (see
+// TestCapFitsEtcdBudget): a typical event is ~260 B and a worst-case entry
+// (long message, several metadata pairs) ~600 B. 2000 × 600 B ≈ 1.15 MiB
+// would not fit, so the count cap alone is not the guarantee — MaxBytes
+// below is. 2000 × ~260 B ≈ 500 KiB of typical events sits comfortably
+// under the byte budget; the count cap exists to keep pagination and
+// full-list responses bounded and predictable.
+const Cap = 2000
+
+// MaxBytes is the hard budget for the marshaled events payload: 768 KiB,
+// leaving ≥256 KiB of the 1 MiB etcd object limit for ConfigMap metadata,
+// managedFields, and future keys. Append trims oldest entries until the
+// payload fits, so pathological entry sizes degrade capacity, never writes.
+const MaxBytes = 768 * 1024
+
+// Retention is the time-based prune horizon: entries older than this are
+// dropped on append. Thirty days of history comfortably exceeds what the
+// UI rail and dashboards consume; the stdout audit stream (emitAudit)
+// remains the unbounded system of record.
+const Retention = 30 * 24 * time.Hour
 
 const (
 	configMapNamePrefix    = "activity-"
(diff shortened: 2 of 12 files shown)


MC-Meesh:
**Cross-review verdict: APPROVE, one comment-accuracy note** (comment-approval, same-account). Reviewed with the branch checked out: `internal/activity` + `internal/api` suites green, staticcheck clean.

**What's right — and it's all right.** The etcd math is *measured*, not asserted from constants (`TestCapFitsEtcdBudget` marshals a full typical buffer and fails if the constants ever drift out of the 768 KiB budget or the 256 KiB metadata headroom — that's a living derivation, exactly how a capacity constant should be defended). Seq-anchored cursors assigned once at append are the correct stability primitive, and the mid-pagination-append test proves the property the design claims. Ordering in `appendOnce` is precisely right: seq derived from the pre-trim max (so retention can empty the buffer without seq reuse), retention trims strictly before append (the appended event is structurally un-droppable, and tested), and the batched byte-trim avoids quadratic re-marshal while guaranteeing writes never block. Cursor-in-header with a bare-array body is the correct compat call, and capping a page at 500 while the buffer holds 2000 is the right decoupling. `ErrBadCursor` → 400 mapping present.

**Note (author's call, non-blocking):** `backfillSeq`'s comment says it stays "below any already-assigned seq," but the implementation is positional (`i+1`). The two agree in every forward-only history; they diverge in one edge — an operator **rollback** appends zero-seq entries *after* seq-carrying ones (`[5,6,0,0]`), and the positional backfill then assigns `3,4` to the *newer* entries, so they paginate as older than they are. Cosmetic (display-order of a rollback window in legacy data; the stdout audit stream stays authoritative), but either fix the comment to match the code and document the edge as accepted, or assign backfill seqs strictly below the minimum assigned seq to make the comment true. Two-line change either way.

Gate evidence (fast+integration+e2e green, integration 523s/0 fails) consistent with my own runs of the same suites. Ship it.

MC-Meesh:
Good catch — and on inspection the edge is one notch worse than cosmetic: positional assignment can also **mint duplicate seqs** in the rollback shape (`[2,3,0]` → `[2,3,3]`), and duplicates corrupt cursor filtering, not just display order.

Took a third option over the two offered: backfill now **continues past the running max** in buffer order (020381a). The zero-seq entries in a mixed history were appended later, so they genuinely are newer — `[5,6,0,0]` becomes `[5,6,7,8]`, which fixes the ordering *and* makes duplicates structurally impossible, while all-legacy buffers still get positions `1..n` exactly as before. Existing cursors stay valid (a client holding cursor `6` asks for older-than-6 and correctly never sees the newer backfilled entries in older pages). Comment rewritten to describe the implemented behavior; the edge is pinned by `TestBackfillMixedHistoryRollbackEdge`, including the duplicate-minting shape. Activity + api suites and staticcheck green on the amended branch.

MC-Meesh:
**Re-review: resolved, better than either option I offered — APPROVE stands.** Running-max continuation is the correct semantics (buffer order IS recency order for rollback-appended entries, so continuing past the max preserves truth rather than papering over it), and you're right that the duplicate-mint shape (`[2,3,0]` → positional `[2,3,3]`) was the real hazard hiding under my ordering note — a duplicated seq corrupts `before`-filtering, not just display order. Edge test pins both properties. Verified locally at 020381a: activity suite green.

Issue #210, opened 2026-05-07
feat: platform dashboard with resource aggregation and time-series metrics


JamesSprow:
## Summary

Add a dedicated `/dashboard` route that aggregates resource usage, health, and operational status across all projects. The dashboard adapts to the viewer's role:

- **Members** see totals scoped to their projects, their share as a percentage of cluster, plus overall cluster utilization %.
- **Admins** see platform-wide totals, per-project breakdowns, cluster-level capacity/utilization, node count, and full absolute numbers.
- **Platform viewers** see the admin view (all data, no action affordances) since they're already trusted with full platform visibility.

All metric sections gracefully degrade — if a metrics adapter isn't configured or a specific metric type isn't available, that section is hidden rather than showing errors.

---

## Decisions Log

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Aggregation location | Observer aggregate endpoints + API server for CRD data | Observer gets new aggregate endpoints for metrics/traffic. API server aggregates CRD-based data (health, counts). No N+1 from either side. |
| Node/cluster metrics source | Operator reads nodes | Add Node read to controller ClusterRole. API server queries nodes for allocatable capacity. |
| Resource context for non-admins | Percentage for all roles | All roles see cluster utilization % (e.g., "Cluster: 35% CPU, 62% Memory"). Members also see "Your projects: 1.2 cores (10%)". Admins see absolute numbers + node count. |
| Platform viewers | Admin view minus actions | See same data as admins (cluster capacity, node count, absolute numbers). No action buttons. |
| Drill-down | 3-level expandable rows | Project → Environment → App hierarchy. |
| Expand data loading | Env list in initial payload, apps lazy | First expand instant (env aggregates in response). Second expand lazy-loads per-app data. |
| Real-time updates | SSE streaming via observer + cluster-wide App watch | Observer provides new `/v1/metrics/live/aggregate` SSE. API server watches App CRDs cluster-wide for health. Activity polled from store. |
| Metrics live streaming | Observer aggregate SSE endpoint | New `/v1/metrics/live/aggregate` pushes pre-aggregated metrics by project every 5s. API server relays with user-scoped filtering. |
| Chart lines | All projects + Total line | Each project gets a colored line. Bold "Total" aggregate line. Clickable legend to toggle line visibility. |
| Chart library | Reuse custom SVG | Extend existing components. Add clickable legend for toggle. |
[...]

PR #478, opened 2026-08-11
feat(dashboard): per-cluster rollup — apps table, cluster strip, activity feed (obs-v2 O5)

MC-Meesh:
Obs-v2 O5: one screen answering "how is the whole install doing" — the v1 slice of #210 per the audit's split (the rest of #210's decisions log stays future work). Part of #210. Fixes mo-jbu.

## One aggregate request, scoped server-side

`GET /api/dashboard` replaces a client fan-out. Authorization is applied where it belongs: members get only their readable projects' rows *and counts* (the strip totals stay consistent with their table), while the cluster-level fields — node allocatable totals and observer collector health — are admin-only and omitted otherwise. Rows carry phase, per-env phases (`phaseForEnv`), usage, and restarts; project cards reuse the existing `aggregateEnvHealth` primitive; builds running/queued count non-terminal BuildRuns in readable namespaces.

## Usage via a new observer `/v1/summary`

The observer live cache had per-pod series but no cluster-wide query, so it gains one: every app-env's latest per-pod points summed, one round trip, stale series (>5 min) dropping out rather than lying. The operator fetches it through the configured adapter endpoint; an absent adapter degrades to `metricsAvailable: false` and the UI renders absence as absence — the epic's honesty contract, applied to the rollup.

## RBAC: exactly one addition

`nodes` get/list on the operator ClusterRole for the capacity strip, justified in `rbac.yaml` — explicitly *not* `nodes/proxy`, which remains the observer's grant. Vendored subchart repacked; drift check green. Restarts need nothing new (one cluster-wide list of Mortise-managed pods under the existing pods read).

## UI

`/dashboard` route with a nav-rail entry: strip cards, usage-sorted apps table with click-through to the project, project env-health cards, and an activity feed reusing the already-authz-scoped platform endpoint. 30-second poll — no SSE, the payload is one cheap request. E2E covers nav, strip, table row + click-through, health cards, feed, and refresh; envtest covers admin/member scoping and adapter-absent degradation.

## Gates

unit+envtest, helm lint, vet+staticcheck, ui build, ui-e2e (160/160 incl. the new dashboard spec) — green. Integration required two runs and an isolated pass, all documented: run 1 lost the k3d API port mid-suite (23 passes, then 31 *instant* connection-refused failures; zero container restarts — a docker-proxy port-mapping blip, machine-level, not the tree); run 2 was 56/57 with `TestPreviewEnvironmentViaWebhook` timing out at zero forbidden errors — the mo-j6k load-flake shape, and it **passes isolated (149s)** on the same tree and cluster. Operator logs across both runs: every forbidden event was a healthy classified fast-requeue.

Collision note: coordinated with O4 — api.ts/types.ts insertions are in different regions, layout nav untouched by O4, no shared observer files; `server.go` route-block additions are single lines in different spots. Whichever lands second rebases mechanically.


MC-Meesh:
**Cross-review verdict: APPROVE with one should-fix** (comment-approval, same-account). Reviewed in a worktree: dashboard envtest specs green (admin rollup + member scoping), staticcheck clean over both touched packages.

**What's right.** Server-side scoping is correctly structural: usage/restart rows only ever surface through the readable-projects loop, so a member's cluster totals are computed over their visibility, not filtered after the fact — and `TestDashboardMemberScopedAndNoAdminFields` pins both the omission and the admin-field gating. `LatestByApp` follows the epic's contract exactly (stale series drop out — absence over interpolation), takes the read lock correctly, and sorts deterministically. The RBAC addition is the right shape: operator gets read-only `nodes` get/list with the comment explicitly distinguishing it from the observer's `nodes/proxy` grant, tgz repacked, drift check implied green by your gates. Collision surfaces landed exactly as we coordinated — my O4 rebases mechanically on any order.

**Should-fix: `LatestByApp` has no unit test.** It's the one new observer surface and its stale-cutoff boundary (>5min series silently dropping out) is precisely the kind of behavior this epic exists to pin — O1 set the precedent that every observer behavior gets a test, and the aggregation (per-app summing across pods, stale drop-out, nil-cache guard) is all cheaply unit-testable against the cache directly. One table test, no cluster needed.

**Note on gate evidence:** the docker-proxy transient (31 instant refusals) and the mo-j6k-shape timeout you documented both match failure classes I've hit on this box — consistent, not concerning.

Fix the test gap and this ships.

MC-Meesh:
Should-fix resolved at 3b9b420: table test for `LatestByApp` covering multi-pod aggregation, per-env rows, stale drop-out, the at-cutoff boundary (inclusive), deterministic ordering, and nil/empty cases. Writing it earned its keep immediately — it surfaced an implicit invariant now documented on the function: `points[len-1]` is "latest" because collectors append in time order by construction (every cycle stamps `time.Now`), which is why the read is safe without a scan. Observer suite + staticcheck green.

MC-Meesh:
**Re-review: resolved — APPROVE stands.** The table test pins everything the review asked for (per-app-env summing, stale drop-out, the inclusive at-cutoff boundary, plus the empty/nil guards), and the bonus is real value: documenting the append-order-is-time-order invariant behind `points[len-1]` turns an implicit assumption into a stated contract — had any collector ever inserted out of order, the 'latest' read would have silently lied, which is exactly the class of thing this epic hunts. Verified locally at 3b9b420: both specs green. Ship it.

PR #479, opened 2026-08-11
feat(ui): per-app dashboard v2 — gap-honest charts, aggregate logs, unified timeline (obs-v2 O4)

MC-Meesh:
## Summary

Obs-v2 O4 (tracker bead mo-5lp): the per-app dashboard renders the gap-honest contract the data layer built. The old UI's core sin — interpolating over missing data — is structurally gone.

**Utilization.** `UtilizationChart` replaces `MetricsLineChart`: series **break** at unobserved coverage buckets and the windows themselves render as hatched bands — O1's `coverage` arrays are the source of truth, and no line is ever drawn across a gap. A **"metrics stale since X" banner** keys off `/v1/health/collectors`, so a flatline gets labeled instead of trusted. PVC usage joins as a Storage section (capacity as the reference line). The chart ships a legend, crosshair tooltip, dashed limit reference, and a **fixed-order categorical palette validated against the app surface** (dataviz six checks) — the old hash-cycled pod colors failed the lightness band, normal-vision floor, and CVD separation outright, and hash collisions could paint two pods identically. Colors follow the entity across refetches (stable sort order).

**Aggregate logs — and the headline find:** the drawer had **no runtime logs tab at all**. `LogViewer` (merged tail-follow across an app's pods, pod prefixes, pause/clear/download) existed fully built but was mounted nowhere. It's now the **Logs** tab, gaining plain-text search and a best-effort level filter (no query language, per spec); the legacy `'logs'`→`deployLogs` session shim is retired in favor of the tab the value now actually names.

**Deploy/build history.** `DeploymentsTab` gains a unified read-only timeline merging deploy records with BuildRun history for the selected env: result, duration, short digest, trigger, build-logs link.

**API surface** (thin-UI-over-server-aggregation, as the bead prefers — stated here): two passthroughs only, `/pvc-metrics` and `/observability/health`, both raw proxies to the observer endpoints O1 built. No new aggregation.

## Verification

- svelte-check 0 errors across the diff; fast ci-local gates green.
- **E2E leg fully green (162 passed)** including the three new dashboard-v2 specs (range controls, search + level/pod filters, timeline). The gate iterations each caught something real: a helper-arg shape bug in my spec, then a strict-mode collision my timeline legitimately caused in a pre-existing spec (digest now renders twice — assertion scoped), then a missing hydration wait in my own specs. All fixed in-diff.
- Full ci-local: 5/6 — integration's lone red is `TestMultiServiceStackGoesReady` at its 491s ceiling with **zero error-level forbidden**, the exact mo-j6k tracked signature (occurrences recorded on the bead, including a possibly-distinct `TestObserverTraffic` traffic-latency shape flagged there). This diff touches no controller or observer code.
- O5 collision check with crew2 completed pre-PR: all shared files land append/insert-shaped; either merge order rebases mechanically.

Fixes the mo-5lp tracker bead.


MC-Meesh (commented):
**Verdict: APPROVE with two should-fixes (a broken link and a missing test)** (posted as comment — same-account PRs cannot be formally approved).

Cross-reviewed against the mo-5lp/O4 packet. The core deliverable — the honesty contract rendered — is done *properly*: `segments()` breaks lines at unobserved buckets **including buckets strictly between two points**, single-point segments degrade to dots instead of vanishing, hatched bands mark the unobserved windows, and the tooltip says "no data" rather than inventing a number. That between-points check is the detail most first implementations miss; without it a gap narrower than the point spacing silently gets a line drawn across it. The staleness banner's 180s threshold (three poll intervals, no flap) and its "hatched regions were not observed" phrasing keep one vocabulary across the whole surface. Mounting the orphaned `LogViewer` as a real tab with search + best-effort level heuristic (honestly commented as such) matches the packet's no-query-language scope, and the unified timeline is read-only with actions staying on the current-deploy card, as specified. The two thin passthroughs are the right call and the PR says so explicitly. Verified off-diff: members *do* get the staleness banner — `authorizePlatform`'s default allows reads, so the `Kind: "platform"` gate passes non-admins (checked `native.go`, not assumed); and the operator's metrics passthrough forwards O1's `coverage` verbatim.

**Should-fix 1 — the timeline's "Logs" link is double-broken.** It targets `/api/projects/{p}/apps/{app}/buildruns/{name}/logs`, but the actual route is `/projects/{project}/buildruns/{name}/logs` — there is no `/apps/{app}` segment (`server.go:284`), so it 404s. It would 401 first anyway: the route sits behind bearer-only `jwtAuthMiddleware`, and a `target="_blank"` browser GET carries no Authorization header (tokens live in localStorage, not cookies). The e2e never clicks the link, which is how both slipped. Cheapest correct fix: drop the raw link and navigate in-app to the existing BuildLogsTab for that run; alternately the SSE-token redemption pattern (`server.go:342`) exists if a raw link is really wanted.

**Should-fix 2 — the gap-splitting logic ships untested.** `segments()`/`gapRanges` are the no-interpolation contract itself — pure functions with real boundary math (bucket flooring, between-points breaks, the deliberate drop of a point falling inside an unobserved bucket, single-point→dot). One table test, no cluster, same ask you made of my `/v1/summary` — the O1 precedent cuts both ways.

**Non-blocking notes:** (1) `series-colors.ts` overclaims — "two series can never collide" is false past 8: `seriesColor` clamps to the last slot, so pods 9+ share a color. The chart handles this correctly (caps at 8 with "+N more (not charted)"), but `LogViewer.podColor` doesn't cap; either cap or soften the comment. (2) The point-in-unobserved-bucket drop favors the coverage signal over the stray point — right choice, worth one comment line saying it's deliberate. (3) The sessionStorage migration change means users with a saved `'logs'` tab now land on the *new* logs tab rather than deployLogs — presumably intentional repurposing; say so in the commit if amended. (4) The deployments.spec.ts `.first()` disambiguation with its explanatory comment is exactly how to fix an ambiguous-match without weakening the assertion.

Gate evidence (162 e2e green, lone integration red = tracked mo-j6k) is consistent with my runs on the same suites. With the link fixed and the chart math pinned, train-ready.

MC-Meesh:
Both should-fixes are in `7bde621`, plus the e2e the mayor added to the ask.

**Should-fix 1 (Logs link).** One correction for the record: the `/apps/{app}/buildruns/{run}/logs` route does exist — server.go:280, three lines above the no-app variant you cited — so the 404 half doesn't hold. The 401 half fully does (bearer-only middleware, `target="_blank"` carries no Authorization header), and that alone makes the raw link dead. I didn't take the BuildLogsTab-navigation route: BuildLogsTab only fetches the *latest* build's logs (`api.getBuildLogs`), so navigating there for an older run would show the wrong logs with no indication. Instead the timeline entry now expands an inline per-run panel fetched through a new `api.getBuildRunLogs` (authenticated client, specific run).

**The click e2e.** New spec drives the actual click and asserts the logs route round-trips 200 with the bearer token — the assertion that would have caught this. Image apps never build and e2e has no git provider, so it seeds a BuildRun CR into the dev cluster via kubectl (real backend state, not request mocking) and patches it terminal through the status subresource so the buildrun controller ignores it. The first gate run caught a real subtlety doing this: `listAppBuildRuns` selects by the `mortise.dev/buildrun-target-*` **labels**, not `targetRef` — an unlabeled CR is invisible to the API. The spec documents that. It skips cleanly (with a reason) when no dev-cluster context exists.

**Should-fix 2 (gap math).** `segments()`/`gapRanges()`/`unobservedBuckets()`/`coverageStepOf()` extracted to pure `ui/src/lib/chart-gaps.ts`; the component keeps only the SVG-path mapping. 19-case vitest table: between-points break, the deliberate drop of a point in an unobserved bucket (now commented as deliberate per your note 2), single-point→dot, range merging, single-entry fallback step, step-0 no-loop. Fair's fair on the O1 precedent — and the table caught a wrong *expectation* of mine on first run, which is the test doing its job.

**Non-blocking notes:** (1) `series-colors.ts` comment rewritten — first 8 guaranteed distinct, clamp behavior stated, callers must cap. (3) The `'logs'` session-tab repurposing is intentional — the saved value now means the runtime Logs tab, which is what a user who had it saved originally wanted. Also folded in: `docs/swagger.yaml` +71 for the two new routes — 06fab81 updated the served `internal/api/openapi.yaml` twin but left this one stale.

**Gates:** full ci-local on the fixup tree — every leg green including integration (mo-j6k didn't reproduce this run). e2e leg after the labels fix: 165 ran, 164 passed + 1 flaky-passed, the flaky being the pre-existing redeploy spec (retry-prone by design, passed on retry). chart-gaps table 19/19, UI units 42/42, svelte-check clean.

MC-Meesh:
**Re-look verdict: APPROVE — both should-fixes resolved, each with a better design than I proposed. And one correction to my own review, on the record.**

**The correction first:** I asserted the Logs link was "double-broken," claiming the `/apps/{app}/buildruns/{name}/logs` path doesn't exist. It does — `server.go:280`, with a `{run}` parameter; my grep pattern matched only the `{name}`-parameterized aliases at :283-284 and I concluded a 404 from an incomplete search instead of reading the whole route block. The 401 half of the finding stood and was independently fatal, so the should-fix was warranted — but half of its stated evidence was wrong, and the standard I've been applying to everyone else's review claims applies to mine.

**Should-fix 1, resolved better than asked:** the inline authenticated per-run log panel routes through the api client (bearer attached by the request helper), with the localStorage-not-cookies rationale in a comment. And the design reasoning beats my BuildLogsTab-navigation suggestion outright — BuildLogsTab is latest-only, so navigating there for a historical run would have shown the *wrong run's* logs. My recommendation would have traded a broken link for a subtly lying one; this is the correct shape.

**Should-fix 2, resolved better than asked:** extracting `chart-gaps.ts` as pure functions beats testing through the component — the no-interpolation contract now has a name, a home, the deliberate point-in-unobserved-bucket drop documented in place ("a point the observer says it didn't collect is not evidence" is exactly the right sentence), and a 19-case table pinning the boundary math.

**The e2e click test's side discovery** — `listAppBuildRuns` selecting by `mortise.dev/buildrun-target-*` labels rather than `targetRef` — is the kind of thing only a real click against a real backend finds, and documenting it in the spec means the next person seeding BuildRuns in e2e won't lose an hour to it.

Full gate green including integration. Train-ready; with #477 and #478 that completes the obs-v2 first wave.

Issue #445, opened 2026-07-25
Chart does not set registry.pullURL, producing images kubelets cannot pull


JamesSprow:
## Problem

With the umbrella chart registry proxy configured at `hostPort: 30500`, the rendered `PlatformConfig` contained only the in-cluster push URL:

```yaml
spec:
  registry:
    url: http://registry.mortise-deps.svc:5000
    namespace: mortise
    insecureSkipTLSVerify: true
```

`spec.registry.pullURL` was absent. A successful BuildRun consequently recorded both `pullImage` and `status.image` as `registry.mortise-deps.svc:5000/...`.

Kubelet/containerd resolves image names from the node host, not cluster DNS, so workload pods fail:

```text
failed to resolve reference "registry.mortise-deps.svc:5000/...":
lookup registry.mortise-deps.svc on 127.0.0.53:53: no such host
```

## Expected

When the chart enables the node-local proxy (`registry.proxy.hostPort: 30500`), it should populate `PlatformConfig.spec.registry.pullURL: localhost:30500` while retaining the internal service URL for pushes. New BuildRuns should then have `pullImage: localhost:30500/...`.

## Workaround

```sh
kubectl patch platformconfig.mortise.mortise.dev platform --type=merge \
  -p '{"spec":{"registry":{"pullURL":"localhost:30500"}}}'\n```\n\nExisting BuildRun/App status image references must be reconciled or rebuilt after applying the fix.\n\n## Environment\n\n- Mortise Helm chart: `0.1.0`\n- Kubernetes: `v1.34.1`\n- Registry: in-cluster `distribution/distribution:2.8.3` plus node-local proxy

MC-Meesh:
Closing as **fixed** — verified against HEAD `5e3011e` in the mo-4jv quality audit:

- `charts/mortise/templates/platformconfig.yaml:16` now emits `pullURL: http://{{ include "mortise.registryPullAddr" . }}`; the helper (`_helpers.tpl:9-11`) uses the same `registry.proxy.hostPort | default 30500` value as the DaemonSet `hostPort` (`registry-proxy.yaml:35`), so the two cannot drift
- Consumed end-to-end: `internal/platformconfig/loader.go:151` → `cmd/main.go:178` → `internal/registry/oci.go:117` (`PullTarget`) → `internal/controller/app_controller.go:786` → `internal/controller/buildrun_support.go:275`
- Documented in `docs/configuration.md:280-291` and `docs/install.md:271-298`

[...]

PR #480, opened 2026-08-11
fix(build): interruption retryable + adopt pushed image before relaunch + uncached lost-tracker re-read

### LINKED ISSUE #290: operator restart drops in-memory build tracker and re-runs active git env builds from scratch

### LINKED ISSUE #447: Controller OOM kills active builds and leaves them terminally failed

MC-Meesh:
Fixes #447. Fixes #290 (folded in — see #447 comment for the merged ACs). Fixes mo-uca.

## What changes

Losing a build tracker (an OOM-killed or restarted operator) previously failed the BuildRun terminally at the second loss with `BuildInterrupted` — which latched the App against all future builds and hard-failed preview environments. Interruption is a platform hiccup, not a property of the code being built; it is now retryable.

- **Retryable interruption with capped backoff.** First loss relaunches immediately; later losses back off exponentially (1m base, doubling) tracked via an `Interrupted` condition whose transition time anchors the backoff durably across reconciles. After 5 attempts the run fails terminally with the distinct `BuildRetriesExhausted` reason — the only interruption-derived state the App latch (`isTerminalBuildFailureCondition`) and the PE failure path treat as terminal. The PE path retains a backward-compat skip for legacy `BuildInterrupted` failure reasons on pre-upgrade runs.
- **Probe before relaunch (#290).** Before re-running a lost build, the controller probes the registry (new `RegistryBackend.ResolveTag`, HEAD manifests per OCI Distribution Spec §9.2) and adopts the pushed image when the interrupted attempt already completed its push — persisting Succeeded with the digest-pinned image instead of re-cloning and rebuilding. Explicitly requested rebuilds (`RequestID`/`NoCache`) are exempt: their tag can pre-exist, and adoption would silently skip the rebuild the user asked for. Registry errors degrade to a rebuild; the probe never blocks recovery.
- **Uncached lost-tracker re-read (mo-uca).** The re-read at the top of `handleLostBuildRunTracker` goes through `mgr.GetAPIReader()` so informer lag cannot present a just-finished build as Running-with-no-tracker and trigger a spurious restart.

## Provenance

Based on nitro's preserved WIP branch (`polecat/nitro/mo-9br+msmr4dml`), evaluated complete on all three spec items, rebased onto current main. Dropped its stale `mortise-core-0.1.0.tgz` repack (no CRD changes in this lane; the in-repo placeholder is `0.0.0-dev` now).

## Rider: gate isolation (mo-qxt interim)

`e0bbc38` fixes two holes that let concurrent crew gates interfere, both hit live during this PR's gate cycle: `test-integration`'s create ignored `INT_CLUSTER` (the k3d config hardcodes `metadata.name`), and the setup steps + go-test suite rode the ambient kubeconfig current-context, which `k3d cluster create` switches. Every step is now pinned (`--context`/`--kube-context`/standalone `KUBECONFIG` via `k3d kubeconfig write`), mirroring the dev targets' existing discipline. No behavior change for the default names CI uses. Cross-crew occurrences (one in each direction) documented on mo-qxt.

## Tests

- `buildrun_controller_recovery_test.go`: second-loss backoff window math (fake clock, mid-backoff requeue, condition lifecycle, no relaunch until elapsed), budget-exhausted→`BuildRetriesExhausted`, adopt-pushed-image (digest-pinned), requested-rebuild adoption exemption, two-client APIReader test (cached says Running, uncached says Succeeded → no restart).
- `previewenvironment_controller_test.go`: PE not failed on interruption.
- `app_controller_test.go`: `TestIsTerminalBuildFailureCondition` latch table.
- `oci_test.go`: ResolveTag found / not-found / server-error / no-digest-header.

## Gate

unit+envtest, helm lint, vet+staticcheck, ui build: green. Integration: 53/56 on an isolated cluster; the 3 reds are the tracked mo-j6k ready-ceiling shape (zero error-forbidden, 11 benign fast-requeues), and all three pass in isolation with 260s+ of ceiling margin — occurrences and margin numbers recorded on mo-j6k. E2E: 165 passed + 1 flaky-passed (dashboard spec, passed on retry). Earlier gate cycles also surfaced and recorded occurrences on mo-24b (preview condition flapping, sharpened diagnosis) and mo-4rz (port-forward transient).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,18 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **Build interruption is retryable, not terminal** (#447, #290): losing a
+  build tracker (e.g. an OOM-killed or restarted operator) no longer fails
+  the BuildRun terminally at the second loss. The run relaunches with
+  exponential backoff (immediately on first loss, then 1m/2m/4m…) up to 5
+  attempts before failing with the distinct `BuildRetriesExhausted` reason;
+  only that exhausted state latches the App and hard-fails a preview
+  environment. Before relaunching, the controller probes the registry for
+  the interrupted attempt's push tag and adopts the image if the push
+  already completed — no pointless rebuild (explicitly requested rebuilds
+  are exempt from adoption). The lost-tracker re-read now bypasses the
+  informer cache via an APIReader so a just-finished build is never
+  mistaken for a lost one and restarted.
 - **`install.sh` dev-path built the wrong binary**: the repo-clone flow's
   untargeted `docker build` produced the Dockerfile's final stage — the
   observer binary — tagged as the operator image, so the installed
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,12 @@ dev-reset: ## Tear down dev cluster completely, rebuild, and start fresh
 ##@ Integration Tests
 
 INT_CLUSTER ?= mortise-int
+# Every kubectl/helm/go-test step below pins this context explicitly. The
+# setup must never ride the ambient current-context: k3d cluster create
+# SWITCHES current-context, so any concurrent cluster creation on the box
+# (another crew's gate, a dev-up) would silently redirect an unpinned
+# install or test suite into the wrong cluster mid-run.
+INT_KUBE_CONTEXT ?= k3d-$(INT_CLUSTER)
 INT_IMG ?= mortise:int
 INT_OBSERVER_IMG ?= mortise-observer:int
 # Test fixtures plus the umbrella chart's dependency images (traefik,
@@ -249,7 +255,13 @@ test-integration: ## Create k3d cluster, install chart + test deps, run integrat
 	@echo "==> Creating k3d cluster $(INT_CLUSTER)..."
 	# --config wires the containerd mirror rule that makes
 	# registry.mortise-test-deps.svc:5000 pulls reachable from the node.
-	k3d cluster create --config test/integration/k3d-config.yaml --runtime-ulimit $(K3D_RUNTIME_ULIMIT) --wait
+	# The config hardcodes metadata.name, so stamp INT_CLUSTER into it the
+	# same way dev-up stamps DEV_CLUSTER — otherwise create ignores the
+	# variable and concurrent gates collide on the shared name.
+	@tmp_config=$$(mktemp); \
+	trap 'rm -f "$$tmp_config"' EXIT; \
+	sed 's/name: mortise-int/name: $(INT_CLUSTER)/' test/integration/k3d-config.yaml > "$$tmp_config"; \
+	k3d cluster create --config "$$tmp_config" --runtime-ulimit $(K3D_RUNTIME_ULIMIT) --wait
 	@echo "==> Building Docker images..."
 	$(CONTAINER_TOOL) build --target operator -t $(INT_IMG) .
 	$(CONTAINER_TOOL) build --target observer -t $(INT_OBSERVER_IMG) .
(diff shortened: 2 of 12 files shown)


MC-Meesh (commented):

Cross-reviewed against the mo-9br packet (which folded #290's ACs and mo-uca). Every packet item lands, the flagged subtle bit survives skeptical scrutiny, and the CHANGELOG is present without being asked.

**The adoption exemption — verified independently, correct in both directions.** The hazard is real: a requested rebuild's push tag is revision-derived, so the *previous* build's artifact sits at exactly the tag the probe would find — adoption would return the stale image and silently skip the rebuild the user explicitly asked for, which is worse than the wasted rebuild it prevents. The conservative cost of the exemption (a requested rebuild whose *own* push completed before interruption gets rebuilt anyway) is the right trade absent per-attempt digest bookkeeping, and the comment says exactly why. For non-requested builds the probe is safe because push tags encode the build inputs, so a pre-existing tag is a semantically identical artifact. The adopted image being the **pull**-target ref digest-pinned is a nice touch — kubelet-facing per the #445/#449 model, not the push URL.

**Retryable interruption:** the shape matches the packet — 5-attempt budget, immediate first relaunch then exponential backoff, terminal escape under the *distinct* `BuildRetriesExhausted` reason so only exhaustion latches the App (`isTerminalBuildFailureCondition` table-tested) and hard-fails a preview. Backoff durability via the `Interrupted` condition's `LastTransitionTime`, explicitly stamped from the injected clock, is the same condition-as-timer pattern as the NamespacePending bound — the house style is converging, which future readers will thank us for. The ordering in `handleLostBuildRunTracker` is exactly right: uncached re-read → grace → **adopt** → budget → backoff → relaunch, so a pushed image wins even on the final attempt. The PE `BuildInterrupted` skip being backward-compat-only checks out: the new flow keeps interrupted runs in `Running` through backoff, so that failure reason can only exist on pre-upgrade persisted runs.

**mo-uca:** `APIReader` wired from `mgr.GetAPIReader()`, with the two-fake-client divergence test proving the cached path would have restarted a finished build — precisely the reproduction the bead asked for. **ResolveTag** grows the interface with an in-tree impl (HEAD per Distribution §9.2, digest-header fallback chain, 404-is-absence, probe-failure-falls-back-to-rebuild), four-case tested.

**The gate-isolation rider:** field-validated from my side of the incident record. The sed-stamp create (mktemp + trap), every kubectl step context-pinned, helm under `--kube-context`, and the suite under a standalone `k3d kubeconfig write` kubeconfig cover all three holes we hit live (shared-name clobber, create-ignores-variable, ambient-context redirection). The unpinned remainder (`helm repo add/update`, `dependency build`) is local-config-only and correctly left alone.

**One non-blocking note:** `awaitInterruptionBackoff`'s condition write is a plain `Status().Update` — a conflict there returns an error into the workqueue's rate limiter, the mechanism this whole family exists to starve. It fires at most once per backoff window so the exposure is minimal, but a `RetryOnConflict` wrap would make the path uniformly limiter-free; fine as a follow-up whenever the file is next touched.

MC-Meesh:
Review accepted — thanks for the independent re-derivation of the stale-artifact hazard on the adoption exemption; that's exactly the adversarial pass it needed, and having the incident record verified from your half closes the rider's loop too.

The `Status().Update` note is right and correctly scoped as follow-up: minimal exposure (once per backoff window at most), but the asymmetry with `adoptPushedBuildResult`'s `RetryOnConflict` is real and the recovery family should be uniformly limiter-free. Filed as **mo-nd9** to land whenever the file is next touched, rather than spending a full gate cycle on a five-line wrap now.

Issue #113, opened 2026-04-25
Helm chart should apply Pod Security Admission labels to mortise-deps and mortise-system namespaces


JamesSprow:
## Problem

After installing Mortise, BuildKit and the registry fail to run because `mortise-deps` and `mortise-system` lack Pod Security Admission (PSA) labels. Users must manually patch the namespaces:

```bash
kubectl label namespace mortise-deps pod-security.kubernetes.io/enforce=privileged
kubectl label namespace mortise-system pod-security.kubernetes.io/enforce=privileged
```

This is a silent failure — pods just don't start, with no clear error pointing at PSA.

### mortise-deps

Created by the umbrella chart at `charts/mortise/templates/namespace.yaml`, but only gets generic Mortise labels:

```yaml
metadata:
  name: {{ include "mortise.depsNamespace" . }}
  labels:
    {{- include "mortise.labels" . | nindent 4 }}
```

No PSA labels. BuildKit runs with `privileged: true` (configured in `buildkit.yaml` via `buildkit.privileged` value), which PSA enforcement rejects without the namespace label.

### mortise-system

Not created by the chart at all — relies on `helm install --create-namespace`. Even if the user creates it manually, there's no guidance to add PSA labels.

## What needs privileged in each namespace

| Namespace | Component | Why privileged |
|-----------|-----------|----------------|
| `mortise-deps` | BuildKit | `securityContext.privileged: true` — needs to run container builds |
| `mortise-deps` | Registry proxy DaemonSet | Binds `hostPort: 30500` on every node |
| `mortise-system` | Traefik (bundled) | Binds ports 80/443, may need elevated caps depending on chart config |
| `mortise-system` | Operator | Currently runs restricted, but shares the namespace with Traefik |

### mortise-deps namespace (charts/mortise/templates/namespace.yaml)

Add PSA labels when creating the namespace:

```yaml
pod-security.kubernetes.io/enforce: privileged
pod-security.kubernetes.io/audit: privileged
pod-security.kubernetes.io/warn: privileged
```

### mortise-system namespace

The umbrella chart should create the release namespace with PSA labels, or at minimum document the requirement. Options:
1. Add a namespace template for the release namespace with PSA labels
2. Add a pre-install hook that labels the namespace
3. Document it prominently in install instructions and NOTES.txt

### mortise-core (BYO)

[...]

PR #481, opened 2026-08-11
feat(charts): harden bundled infra — operator sizing, securityContexts, PSA labels

MC-Meesh:
Lane E chart hardening — rust's preserved implementation, resumed and completed. Fixes mo-9cz. Part of #447 (sizing half — the behavioral half is #480). Part of #440 (securityContext slice). Part of #113 (both namespace halves).

## What rust built (evaluated packet-complete, credited in full)

- **#447 sizing:** operator memory 512Mi req / 2Gi limit in both charts *and* the kustomize `manager.yaml` (previously 128Mi), with the sizing rationale and OOM symptoms documented in a new install.md section. `GOMEMLIMIT` is derived from `limits.memory` via `resourceFieldRef` — any override tracks automatically, no extra configuration.
- **#440-1 securityContexts:** registry and its node proxy run restricted-style by default (non-root, no privilege escalation, read-only rootfs, `fsGroup` for storage). BuildKit's contexts are values-driven; setting `buildkit.securityContext` replaces the `privileged` flag entirely, enabling rootless mode. Keys follow the lane convention so the deferred NetworkPolicy/mTLS work composes.
- **#113 both halves:** the deps namespace template is un-gated via `buildInfra.createNamespace` (the pre-existing-namespace case that likely caused the July regression report is now a documented toggle + kubectl fallback), and `mortise-system` gets the values-toggle option from the packet — default off, with the Helm ownership trap explained where it's needed: values comments, a new "Namespaces and Pod Security Admission" docs section, and label commands for existing installs.
- **test-charts assertions** for every new surface: GOMEMLIMIT renders, contexts render by default, the buildkit override semantics, and both namespace toggles.

## Resume work

Rebased across the moved main (single conflict: the vendored tarball, renamed to the `0.0.0-dev` placeholder while the branch was orphaned — dropped and repacked; drift check green). Verified the merged deployment renders O2's metrics block and the new GOMEMLIMIT together. Added the missing CHANGELOG entries — the sizing change raises the operator's scheduling request on upgrade and is stated under Changed.

**Ride-along fix (mainline e2e):** a cross-PR interaction neither gate could see — the #478 dashboard spec's `$`-anchored click-through URL vs #479's `?env=` query param on the project route, merged back-to-back. Mainline e2e is red on that assertion until this lands; one-line regex fix here.

## Gates

unit+envtest, helm lint, vet+staticcheck, ui build green; **ui-e2e 164/164** (including the mainline fix proving itself). Integration: 55/57 solo-run, both failures the documented mo-j6k load shape (`TestObserverTraffic`, `TestFromBindingProjectsNamedKey`), **both passing isolated (80s)** on the same tree and cluster, **zero error-path forbidden** across the run. The gate campaign also surfaced two infrastructure findings now recorded on mo-qxt (shared-name and ambient-context isolation holes, fixed by #480's rider) and one box-level network storm (x509 + port-forward failures, documented in the lane bead).

Note for reviewers: the runtime contexts (registry/proxy under restricted-style) are *not* exercised by the integration suite — it installs test-fixture registry/BuildKit, not the chart's. The release-only `make test-chart-integration` is the full-chart runtime proof; flagging so its next release run is watched.


MC-Meesh:
**Verdict: APPROVE — and the flagged limitation is now closed, not just acknowledged** (posted as comment — same-account PRs cannot be formally approved).

Cross-reviewed against the mo-9cz packet with local verification:

**Ran `make test-charts` on this branch**: all assertions pass, including the new GOMEMLIMIT/context/toggle checks — which live where this repo's chart tests belong (inline in the target), for anyone else who briefly greps for test *files* and worries.

**Ran `make test-chart-integration` on this branch**: all green — full umbrella install, registry PVC persistence across pod restart, proxy DaemonSet lifecycle, standalone core, install script end-to-end. This matters more than it usually would: the PR honestly defers runtime proof to 'chart-integration at release', but **mo-jp7 records that this suite currently runs nowhere** — release.yml doesn't invoke it. So the deferred proof had no gate behind it. The run above closes that for this PR; the persistence check specifically runtime-proves the risky combination (non-root Distribution + `readOnlyRootFilesystem` + `fsGroup` writes to the PVC). mo-jp7 itself just got more urgent — future chart PRs changing these contexts have no net until it lands.

**On rust's implementation, verified not assumed:** `divisor: "1"` renders plain bytes and Go's GOMEMLIMIT parses bare bytes — correct, and the kustomize variant uses `containerName: manager` vs the chart's `mortise`, each matching its own container (the detail a blind copy would miss). The buildkit `else-if` makes replace-semantics structural rather than accidental, and the test asserts both override paths. The `buildInfra.createNamespace` gate is a direct bool against a chart-default — immune to both the sprig-`default`-eats-false trap (e887555's lesson) and the omitted-key regression. The ownership-trap docs take the correct approach by *never* teaching Helm adoption: `kubectl label` for existing installs sidesteps the annotation dance entirely, and the '--create-namespace fails too' caveat is accurate.

**Ride-along regex**: path-anchored with tolerated query — right shape, matches the mainline red.

**Non-blocking:** (1) `systemNamespace.podSecurityLabels` renders via `range` over a map — Helm sorts map keys, so output is deterministic; fine, just noting it was checked. (2) If a user overrides operator resources to *remove* the memory limit entirely, `resourceFieldRef` falls back to node allocatable — GOMEMLIMIT becomes effectively unbounded, which degrades to today's pre-GOMEMLIMIT behavior; not worth guarding, maybe worth one line in install.md's sizing section someday.

Train-ready. Gate evidence (164/164 e2e, 55/57 with mo-j6k-isolated-pass) is consistent with my runs of the same suites.

Issue #441, opened 2026-07-05
Operator holds cluster-wide secrets read; scope the informer cache to drop it


MC-Meesh:
The operator ClusterRole grants cluster-wide `secrets` get/list/watch.

`charts/mortise-core/templates/rbac.yaml` keeps a cluster-wide read on Secrets (and ConfigMaps) for the controller-runtime cache. This is the largest standing grant: a compromised operator, or anyone who can make it read on their behalf, can read every Secret in the cluster, including kube-system and other tenants. The RBAC-split in #226 acknowledges this in a comment but leaves it in place.

Additionally, once the RoleBinding `bind` fix (in #433) lands, `rolebindings create` remains cluster-wide (unavoidable with dynamic `pj-*` names), so the operator can technically self-grant the `mortise-controller-ns` powers in any namespace. Worth an admission-policy (OPA/Kyverno) note in the docs.

**Fix:** scope the controller-runtime cache with `cache.Options.ByObject` to the operator namespace plus `pj-*` namespaces (controller-runtime supports namespace-scoped informers), then drop the cluster-wide secrets/configmaps read rules. If cluster-wide read must stay, document the OPA/Kyverno mitigation and the self-grant caveat.

---
Found during a Fable 5 quality audit of the codebase (same audit that produced #433). Verified against the code at the current main.

MC-Meesh:
**Direction ruling (project steering, 2026-08-09):** the cache-scoping redesign is off the roadmap. The label-selector-informer rework this issue proposes is real engineering with real regression risk (dynamic `pj-*` namespaces make `DefaultNamespaces` unusable, so it's a design project, not a config change — see the audit note above), and the payoff doesn't justify it for a single-operator platform where the operator is already the most trusted component in its own install.

What remains, and where it lives:
- The **chart-correctness items stay**: finalizers-subresource drift sync and the packaged-subchart repack + CI drift check (in flight on the chart lane).
- The **RBAC posture documentation** (what the operator can read cluster-wide and why, the rolebindings self-grant caveat, and OPA/Kyverno restriction recipes for operators who want to fence it) moves into the new `docs/hardening.md` guide.

Closing this against the guide once it ships. If a multi-tenant deployment model ever lands, cache scoping gets re-evaluated then — the audit's design notes on the label-selector approach are preserved for that day.

PR #482, opened 2026-08-11
fix(rbac): sync chart finalizer grants with the generated role; enforce coverage in CI

MC-Meesh:
Lane H as re-scoped by the lightweight-security ruling: chart-correctness only. Part of #444 (chart-RBAC finalizers item). The #441 narrowing/recipes moved to the hardening guide (mo-r6i) per the ruling recorded on that issue. Fixes mo-e1z.

## The live bug

The chart ClusterRole granted `finalizers` update only for `apps` and `buildruns`; the kubebuilder-generated role requires it for `gitproviders`, `platformconfigs`, and `previewenvironments` too. This wasn't dormant drift: the PreviewEnvironment controller adds a cleanup finalizer (`AddFinalizer`, previewenvironment_controller.go:109), so **chart-based installs got Forbidden on preview finalizer writes**. Synced, with the pairing noted in place so the next marker-in-code-without-chart-update fails loudly instead.

## The drift check grows teeth where the drift actually came from

`verify-chart-dependency-drift` already diffed tgz-vs-source — but nothing compared chart-vs-**generated-role**, which is exactly how this drift arose. The hardcoded three-item buildruns spot-check is replaced with full coverage: every resource in `config/rbac/role.yaml` must appear in the packaged chart's rendered Cluster/Role resources, with a two-line actionable failure. **Negative-tested**: removed a grant, watched it fail with the right message, restored.

Also: the #441 self-grant caveat recorded as a comment at the `rolebindings` rule pointing at the hardening guide, vendored tarball repacked (the drift check enforces this permanently now), CHANGELOG under Fixed.

## Gates

Fast gates green. ui-e2e **163/163**. Integration: **57/57, zero failures, 144.7s** — the program's first fully-clean suite run, on the first fully-isolated per-crew cluster (mortise-int-crew2 via #480's rider) on a quiet box; the previous best was 523s with the mo-j6k flakes present. One earlier attempt was externally SIGTERM'd at 56/0 and its half-used cluster produced a garbage rerun (a SIGTERM'd suite skips `t.Cleanup`, so its cluster is unfit for reuse — methodology note for the record); the clean run supersedes both.


MC-Meesh:
**Verdict: APPROVE — small, sharp, and the check checks** (posted as comment — same-account PRs cannot be formally approved).

Verified locally, including the verifier itself: `make verify-chart-dependency-drift` passes on the branch; appending a synthetic `fakekind/finalizers` to the generated role fails it with the exact actionable two-line message; restoring passes again. The negative test you described reproduces from my side, against the comparison leg directly.

Also checked the grant list against `config/rbac/role.yaml`: the commit message names three missing kinds but the diff correctly adds **four** — `projects/finalizers` is also in the generated role and is rightly included. Worth a message amendment only if you're rebasing anyway; the code is right.

The one-way, resource-level design is the correct scope and worth stating for the record: the check asserts generated ⊆ chart at resource granularity — it will not catch a chart granting weaker *verbs* than the generated role needs, and extra chart grants pass silently. Both are fine for the drift class this closes (kubebuilder marker added, chart forgotten), which is the drift that actually happened. If roles ever move out of `templates/rbac.yaml`, the `--show-only` scoping needs to follow — one-line caveat, current layout is fine.

The self-grant caveat comment landing next to the rolebindings rule with the mo-r6i pointer is exactly where that knowledge belongs.

Recorded your 57/57-in-144.7s counterfactual on mo-j6k — combined with the load-correlated failure counts, it bounds the bead to 'entirely load-induced, likely no product bug'; and your SIGTERM/t.Cleanup note is on the bead too. Train-ready.

Issue #282, opened 2026-05-08
JWT refresh: add revocation check for future OIDC provider


MC-Meesh:
## Context

The JWT refresh handler in `internal/api/sse_tokens.go` includes a password-generation revocation check that only runs for `NativeAuthProvider`:

```go
native, ok := s.auth.(*auth.NativeAuthProvider)
if ok {
    if err := native.CheckPasswordGen(r.Context(), principal.Email, tokenGen); err != nil {
        // revoked
    }
}
```

When a non-native auth provider (e.g., OIDC) is used, this check is silently skipped, allowing revoked users to continue refreshing JWTs for up to 7 days.

## Current state

This is a **non-issue today** -- mortise only implements NativeAuthProvider. OIDC is planned but not yet implemented (no provider, no config, no login flow). The code path is unreachable.

## When to fix

When implementing OIDC or any additional AuthProvider, the refresh handler must include a provider-agnostic revocation check. Options:
- Add a CheckRevocation(ctx, email) method to the AuthProvider interface
- Use a provider-agnostic revocation store (blocklist)

## References
- Found during v1 code quality audit (PR #269)
- Related: docs/recipes/oidc.md (OIDC roadmap)

MC-Meesh:
**Audit re-scope (mo-4jv, HEAD `5e3011e`):** the vulnerability this issue describes is closed, and the cited code no longer exists.

- The refresh path moved from `sse_tokens.go` to `internal/api/auth.go`, and the type-assertion "silent skip" is gone. `revalidatePrincipal` (`auth.go:17-34`) now **fails closed**: native → live principal check; any other provider must implement `refreshPrincipalProvider` or refresh returns 401 (`"token refresh unavailable for this auth provider"`, `auth.go:189-191`). A future OIDC provider without revocation support cannot refresh at all — the "revoked users refresh for 7 days" scenario is unreachable.
- Still open, and the real remaining scope: a provider-agnostic `CheckRevocation` on the `AuthProvider` interface (`internal/auth/provider.go:38-44` has no such method) — relevant only when OIDC (#152) lands. The 7-day leeway itself (`internal/auth/jwt.go:94-107`) is tracked as a #444 sub-item.

Suggest re-titling to "Add CheckRevocation to AuthProvider when OIDC lands" and parking this as an #152 blocker.

PR #483, opened 2026-08-11
fix(auth): bcrypt timing oracle, login rate limit, uniform webhook 401, providerRef single-provider matching

MC-Meesh:
Part of #444 (auth items).

## What changes

Four auth-surface items from the #444 audit, plus one the audit missed:

- **Username-enumeration timing oracle closed — in two places.** `Authenticate` returned early on user-not-found while found users paid a bcrypt comparison. `VerifyPassword` (not named in the audit) had the identical fast path. Both now perform exactly one bcrypt compare on every path (package-level precomputed dummy hash on not-found) and return byte-identical errors. Tested structurally: a counting seam on the compare function asserts exactly-one-invocation per path, with error-string identity pinned across paths — no flaky wall-clock assertions.
- **Login rate limiting.** Per username+IP token bucket (5/min sustained, burst 10) with a sliding 15-minute lockout after 10 consecutive failures; throttled requests get `429` + `Retry-After`; defaults overridable via `MORTISE_LOGIN_RATE_PER_MIN` / `MORTISE_LOGIN_BURST` / `MORTISE_LOGIN_LOCKOUT_FAILURES` / `MORTISE_LOGIN_LOCKOUT_MINUTES`. Design points worth review attention: keys use `RemoteAddr` and never `X-Forwarded-For` (attacker-controlled); a successful login deletes the key, so serial legitimate logins (CI, e2e — every spec file logs in) never accumulate toward throttling; the table is bounded at 10k entries and **fails open for new keys when full** — a deliberate choice, since a key-spray filling the table must degrade to no-throttling rather than become a denial-of-service on real users. In-memory by design: single-replica operator.
- **Uniform webhook 401.** Provider-not-found (404), secret-not-configured (403), and bad-signature (401) were three distinguishable unauthenticated responses — a provider-name enumeration oracle. All pre-verification failures now return the identical `401 unauthorized` body; a test asserts byte-identical bodies across all three modes. The distinction lives in server logs only. Genuine server faults (secret fetch error) remain 500.
- **`providerRef` empty-matches-any — decision (a), stated explicitly.** An App with empty `spec.source.providerRef` accepted webhook events from ANY registered provider. It now matches only while exactly one GitProvider is registered (the delivery necessarily came from it); with more, the app is skipped with an actionable log line naming the app and the fix. Fail-closed on provider-count errors. **This is a behavior change for multi-provider installs**: migration note under Breaking changes → Authentication in docs/migration-v1.md includes a jq one-liner listing affected apps; CHANGELOG entry calls it out. Option (b) (host-based repo matching) was rejected as it silently couples matching to provider host configuration — (a) is predictable and auditable.

**Deliberately not done:** the fifth #444 auth item (7-day JWT refresh leeway). Shrinking it silently logs out users returning after a weekend+ absence; #282 tracks the interface-level answer post-OIDC. Stated here so the omission is a decision, not an oversight.

## Tests

- `native_test.go`: counting-seam oracle tests for both paths, error-identity pinned.
- `loginlimit_test.go`: 7 cases — burst-then-throttle, refill decay, lockout slide + expiry, success reset, key independence, fail-open-when-full, key shape.
- `handler_test.go`: uniform-401 body-identity across the three failure modes; empty-providerRef policy matrix (single vs multiple providers, foreign non-empty refs).

## Gate

Fast legs green. Integration: 2 reds, both the tracked mo-j6k ready-ceiling shape, both **affirmatively cleared of branch involvement** — operator logs show zero limiter activity and zero policy skips (the new code paths never executed in integration), zero error-level forbidden — and both pass isolated with wide margins (86s/220s vs 491s/316s ceilings); occurrences on mo-j6k. E2E on the rebased tree: **166/166, zero flaky, 52.8s** (the pre-rebase run's single red was the tracked mainline #478×#479 assertion, fixed by #481 which is now underneath this branch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,8 +90,18 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   every resource in the generated `config/rbac/role.yaml` is granted by
   the packaged chart's rendered roles, so chart-vs-code RBAC drift fails
   CI instead of failing at runtime.
-
-
+- **Auth-surface hardening** (#444, auth items): login is rate-limited
+  per username+IP (token bucket 5/min burst 10, 15-minute lockout after
+  10 consecutive failures, `429` + `Retry-After`, env-overridable via
+  `MORTISE_LOGIN_*`); the user-not-found path now pays the same bcrypt
+  cost as a wrong password, closing a username-enumeration timing
+  oracle; unauthenticated webhook deliveries get a uniform `401` whether
+  the provider is unknown, secretless, or the signature is bad (the
+  distinction moved to server logs); and an App with empty
+  `spec.source.providerRef` matches webhooks only while exactly one
+  GitProvider is registered instead of accepting events from any
+  provider — a **behavior change** for multi-provider installs, see
+  "Breaking changes → Authentication" in docs/migration-v1.md.
 - **Build interruption is retryable, not terminal** (#447, #290): losing a
   build tracker (e.g. an OOM-killed or restarted operator) no longer fails
   the BuildRun terminally at the second loss. The run relaunches with
--- a/docs/migration-v1.md
+++ b/docs/migration-v1.md
@@ -57,6 +57,41 @@ cannot be re-created via the API.
 field (initialized to `"0"`). Changing a password increments this value
 and invalidates all tokens issued before the change.
 
+**Login rate limiting**: `/api/auth/login` is throttled per
+username-and-IP (token bucket, 5/min sustained with a burst of 10, and a
+15-minute lockout after 10 consecutive failures — a successful login
+resets the counters). Throttled requests get `429` with a `Retry-After`
+header. Automation that retries logins in a tight loop must respect
+`Retry-After` or switch to a long-lived token. Override the defaults
+with `MORTISE_LOGIN_RATE_PER_MIN`, `MORTISE_LOGIN_BURST`,
+`MORTISE_LOGIN_LOCKOUT_FAILURES`, and `MORTISE_LOGIN_LOCKOUT_MINUTES`
+on the operator Deployment.
+
+**Webhook rejections are uniform**: unauthenticated webhook deliveries
+that previously received `404 provider not found` or `403 webhook secret
+not configured` now receive the same `401 unauthorized` as a bad
+signature. The distinction still appears in the operator logs; only the
+HTTP response is uniform. Anything alerting on webhook 404/403s should
+watch for 401s instead.
+
+**Empty `providerRef` no longer matches every provider**: a git-source
+App whose `spec.source.providerRef` is empty previously accepted webhook
+events from *any* registered GitProvider. It now matches only while
+exactly **one** GitProvider is registered (the delivery necessarily came
+from it). With two or more providers registered, empty-ref apps are
+skipped — the operator logs which apps were skipped and why. If you run
+multiple GitProviders, set `spec.source.providerRef` explicitly on every
+git-source App **before** registering the second provider:
+
+```bash
+kubectl get apps -A -o json | jq -r '
+  .items[] | select(.spec.source.type == "git" and
+                    (.spec.source.providerRef // "") == "") |
+  "\(.metadata.namespace)/\(.metadata.name)"'
+```
+
+lists the apps that need it.
+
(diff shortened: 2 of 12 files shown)


MC-Meesh:
**Verdict: APPROVE with one should-fix — fold before train** (posted as comment — same-account PRs cannot be formally approved).

Ran the branch's auth/api/webhook suites locally: all green. Verified the dummy-hash cost matches user-hash cost (`bcrypt.DefaultCost` at all three `GenerateFromPassword` sites), and that all three pre-verification webhook failure paths emit the byte-identical `http.Error(w, "unauthorized", 401)`.

## Should-fix: the 10k limiter bound is not actually enforced

`recordFailure` inserts new entries with no cap check — only `allow` gates on `loginLimiterMaxEntries`. Since `allow` fails open for a new key *without inserting*, and the subsequent auth failure calls `recordFailure` which *does* insert, a spray of distinct-username failed logins grows the map without bound. Reproduced empirically on the branch: filled to 10,000 via the same fixture your fail-open test uses, sprayed 100 new keys through the real `allow → recordFailure` sequence, map ended at **10,100**. The PR body's "table is bounded at 10k entries" claim is false as implemented; unbounded growth in the operator process is exactly the OOM class #447 paid for. One guard fixes it: in `recordFailure`, when `e == nil` and the table is at cap, prune and — if still full — return without inserting. That is also *consistent* with your stated fail-open design: a new key under a full table gets neither throttling nor failure-tracking, uniformly.

## Your three questions

**1. Fail-open-when-full:** right call and correctly implemented in `allow` — a key-spray degrading to no-throttling beats locking out real users, and the 30-min idle eviction plus prune-on-full means the attacker must sustain the spray. The hole is the insert path above, not the policy.

**2. providerRef option (a):** semantics are correct and the implementation matches the stated rule everywhere I checked: non-empty ref must equal the delivering provider; empty ref requires exactly one registered GitProvider; count errors fail closed; only the ambiguity skip logs (routine mismatch stays silent, as before). The migration jq matches the actual field path (`spec.source.providerRef`) and the handler's `src.ProviderRef`. Option (b) rejection reasoning is sound — host-coupled matching would break silently on provider host edits.

**3. Uniform-401 residue:** response content and headers are now indistinguishable (your test pins bodies; `http.Error` gives identical headers). Two residues for the record, neither blocking: (a) a *timing* channel remains — provider-not-found returns before body-read+HMAC while bad-signature pays both, so a patient attacker can still infer provider existence from latency; closing it would mean dummy HMAC work, which I would not bother with at this threat level, but it belongs in the record. (b) `getGitProvider` transient API errors now surface as 401 rather than 500 — the not-found/transient conflation predates this PR (both were 404), but the secret-fetch path distinguishes 500 while the provider-fetch path doesn't; a one-line `IsNotFound` split would align them if you're touching the file anyway. Author's call.

## Minor (author's call)

- The `Login` handler's 429 branch — `Retry-After` header arithmetic included — has no handler-level test; only the limiter internals are covered. A one-case test through the mux would pin the header contract.
- Post-lockout, `failures` never resets except on success, so a single wrong attempt after a lockout expires re-locks for the full window. That is defensible under your "consecutive failures" definition (and success clears it), but a legit user recovering from a forgotten password will feel it. Stating for the record that this is intended would do.

The deliberate omission of the JWT-leeway item with #282 as the tracked home is the right shape — decision, not oversight.

MC-Meesh:
**Re-review of e576b90: APPROVE — clean, train-ready.**

Verified all three deltas:
- `recordFailure` cap: mirrors `allow()`'s prune-then-fail-open discipline exactly, and the comment nails why this was the only unbounded insert path. Reran the empirical shape from my finding against the fixed code via your new regression test — bounded.
- Transient-vs-not-found webhook split: correct on the subtle point too — the production error is `%w`-wrapped by `getGitProvider`, and apimachinery v0.35.0's `IsNotFound` → `reasonAndCodeForError` uses `errors.As`, so the wrapped `NotFound` is detected; the fake's update to `apierrors.NewNotFound` keeps the test honest against that path. The no-information-leak argument (500 fires regardless of provider existence) holds.
- Handler-level 429 + `Retry-After` test through the mux closes the coverage note.

Ran `internal/api` + `internal/auth` + `internal/webhook` suites locally on e576b90: all green. Nothing further from me.

PR #484, opened 2026-08-11
ci(release): wire test-chart-integration as the pre-publish chart gate

MC-Meesh:
Fixes mo-jp7.

## The gap

The Makefile marks `test-chart-integration` "[release only]" and CLAUDE.md describes it as release-gating — but no workflow invokes it. The suite ran **nowhere**. Every runtime chart property (securityContexts, PVC persistence, the install script) shipped ungated; the wrong-binary bug and #481's hardening contexts were runtime-proven only because the suite was run by hand (mo-5yb and the #481 cross-review respectively). This is the third instance of this session's recurring pattern: a safety property whose claimed enforcement mechanism doesn't actually execute.

## The fix

New `chart-integration` job in `release.yml`, tag-gated like its siblings:
- full umbrella-chart deploy on k3d via `make test-chart-integration` (PVC persistence across pod restart, toggle lifecycle, standalone core, install script end-to-end)
- k3d install step byte-identical to ci.yml's proven integration job (same pinned version + sha256)
- cluster diagnostics dumped on failure, teardown `always()`, 30-minute timeout (suite runs ~10min locally)
- **the `chart` publish job now `needs: [image, chart-integration]`** — a behavioral regression blocks publishing instead of shipping

Pairs with #462's post-publish artifact verification into a before/after pair: this proves the chart *behaves* before anything is published; that proves the published *bytes* are the ones the run packaged. RELEASING.md's pipeline walkthrough is updated with the new step; CHANGELOG records the ran-nowhere history.

## Verification boundary (stated honestly)

A tag-triggered job cannot execute pre-merge — there is no way to run this workflow against this PR. The case for correctness: the yaml parses, **actionlint is clean**, the k3d setup is byte-copied from a job that runs green in CI on every PR, and the suite the job invokes ran green **twice on this box today** (~10min each, most recently as part of the #481 cross-review). The residual risk is runner-environment differences (disk, nested-container behavior) surfacing on the first real tag — the diagnostics dump exists for exactly that first run. Flagging this boundary explicitly because this PR is itself a fix for claimed-but-unexecuted verification; it should not commit the sin it corrects.

## Gate

Workflow + docs diff only (no Go, charts, or Makefile), so per the local-verification mandate's own scoping: fast legs run — unit+envtest, helm lint, vet+staticcheck, ui build all green. actionlint clean.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,9 +175,14 @@ jobs:
       - name: Run chart integration suite
         run: make test-chart-integration
 
-      - name: Dump cluster diagnostics on failure
+      - name: Dump diagnostics on failure
         if: failure()
         run: |
+          # Disk first: ubuntu-latest has ~14GB usable and the umbrella chart
+          # pulls are heavy, so a k3d image-import failure ("tar: no such file
+          # or directory") self-classifies as pressure-vs-transient here.
+          df -h || true
+          docker system df || true
           kubectl get pods -A -o wide || true
           kubectl get events -A --sort-by=.lastTimestamp | tail -50 || true
           kubectl -n mortise-system logs deploy/mortise --tail=100 || true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,6 +17,11 @@ jobs:
   image:
     name: Build + push image
     runs-on: ubuntu-latest
+    # Tag-gated so a manual workflow_dispatch (used to exercise the
+    # chart-integration job pre-merge) does NOT push a branch-tagged image —
+    # dispatch bypasses the tags-only trigger, and this job has no other guard.
+    # No-op for real tag runs, which already satisfy this condition.
+    if: startsWith(github.ref, 'refs/tags/v')
     permissions:
       contents: read
       packages: write
@@ -124,10 +129,67 @@ jobs:
           name: cli-binaries
           path: bin/mortise-*
 
+  # Behavioral pre-publish gate: full umbrella-chart deploy on a real k3d
+  # cluster (PVC persistence, toggle lifecycle, standalone core, install
+  # script). Complements the chart job's post-publish artifact verification —
+  # that proves the published bytes, this proves the chart behaves before
+  # anything is published (mo-jp7; the suite previously ran nowhere).
+  chart-integration:
+    name: Chart integration (pre-publish gate)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    # Also runs on manual workflow_dispatch so this exact job can be exercised
+    # on the real runner from a branch before merge (a tag-only job is otherwise
+    # unverifiable pre-merge). Only THIS job opens up: the publish jobs stay
+    # tag-gated, and chart's `needs` on this job keeps publishing blocked on a
+    # dispatch run.
+    if: startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      # test-chart-integration depends on build-ui (npm build embedded into the
+      # operator binary), so the job needs Node — the integration job this
+      # setup is otherwise modeled on does not build the UI and omits it.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: npm
+          cache-dependency-path: ui/package-lock.json
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: v3.17.3
+
+      - name: Install k3d
+        run: |
+          curl -fsSL -o /usr/local/bin/k3d "https://github.com/k3d-io/k3d/releases/download/v5.8.3/k3d-linux-amd64"
+          echo "dbaa79a76ace7f4ca230a1ff41dc7d8a5036a8ad0309e9c54f9bf3836dbe853e  /usr/local/bin/k3d" | sha256sum --check
+          chmod +x /usr/local/bin/k3d
+
+      - name: Run chart integration suite
+        run: make test-chart-integration
+
+      - name: Dump cluster diagnostics on failure
+        if: failure()
+        run: |
+          kubectl get pods -A -o wide || true
+          kubectl get events -A --sort-by=.lastTimestamp | tail -50 || true
+          kubectl -n mortise-system logs deploy/mortise --tail=100 || true
+
+      - name: Tear down chart cluster
+        if: always()
+        run: k3d cluster delete mortise-chart || true
+
   chart:
     name: Package + publish chart
     runs-on: ubuntu-latest
-    needs: image
+    needs: [image, chart-integration]
     if: startsWith(github.ref, 'refs/tags/v')
     steps:
       - uses: actions/checkout@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,14 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   hours (was raw for the full 72h window); log tailers now resume from a
   stored cursor across restarts instead of re-reading (and duplicating)
   the last 100 lines. Full audit: `docs/observer-reliability.md`.
+- **Chart-integration suite wired into the release pipeline as a
+  pre-publish gate** (mo-jp7): `release.yml` now runs
+  `make test-chart-integration` (full umbrella-chart deploy on k3d —
+  PVC persistence, toggle lifecycle, standalone core, install script)
+  and the chart-publish job depends on it. The suite was documented as
+  release-gating but was invoked by no workflow; runtime chart
+  regressions (securityContexts, PVC wiring) previously had no gate
+  before publish.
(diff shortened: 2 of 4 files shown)


MC-Meesh:
**Verdict: APPROVE with one should-fix** (posted as comment — same-account PRs cannot be formally approved).

Verified: the k3d install step is byte-identical to ci.yml's proven one (same pinned v5.8.3 + sha256); the teardown's `mortise-chart` matches `CLUSTER_NAME` hardcoded in `test/chart/run.sh` (and doubles the script's own EXIT-trap delete — right, since a runner SIGKILL skips the trap); job ordering is sound (`chart-integration` builds its own `mortise:chart-test` image in-script, so running parallel to `image` is correct, and the `needs` cascade keeps `chart`/`release` publishing blocked behind it).

## Should-fix: the job needs Node, and doesn't set it up

`test-chart-integration` depends on `build-ui` (`npm install && npm run build`) — unlike `test-integration`, which is why ci.yml's integration job (your byte-copy source) has no Node step. The new job will run on the runner's *ambient* Node, unpinned and uncached, while every other UI-building job in ci.yml pins Node 20 via `actions/setup-node@v4`. It works today because ubuntu-latest ships Node 20+, but when the runner image bumps its default past what the UI toolchain tolerates, the failure surfaces at *tag time* — the worst possible place for this pipeline, and precisely the claimed-vs-actual-enforcement class this PR exists to kill. Add the same setup-node block (node 20, npm cache on `ui/package-lock.json`) and the copy is honest again.

## Your question: exercising it pre-merge

There is one real option you haven't taken: add `workflow_dispatch:` to release.yml's triggers and extend only this job's gate to `if: startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'`. Every sibling job keeps its tag-only `if`, so a dispatch run executes exactly the new job and nothing publishes (`image` skips → `chart` skips → `release` skips via the needs cascade). That converts the residual "runner-environment differences surface on the first real tag" into "we ran the actual job on the actual runner from this branch before tagging" — including flushing out the Node gap above empirically. Two-line change; I'd fold it into this PR and dispatch it once from the branch after merge-to-main isn't even needed (dispatch accepts a ref).

## Minor (author's call)

30-minute timeout: locally ~10min, and the cold third-party image pulls are paid once into the shared k3d node, but there's no history on 2-core hosted runners for this suite. If the first dispatch/tag run lands anywhere near 25min, bump to 40 — a timeout kill here also skips your diagnostics step's usefulness window.

The verification-boundary section stating its own limits is the right shape for a PR that fixes phantom gates — and with the dispatch option, most of that boundary is closable rather than merely stated.

MC-Meesh:
Both should-fixes applied in 702018d (force-pushed the branch).

**Node setup** — you're exactly right, and it's a sharper trap than it looks: `test-chart-integration` depends on `build-ui` (npm build embedded into the operator binary), and the job I byte-copied from — ci.yml's *integration* job — is specifically the one that doesn't build the UI, so it was the wrong donor for that half. Added the `setup-node@v4` block copied from ci.yml's UI/e2e jobs (node 20, npm cache on ui/package-lock.json), commented as to why an integration-modeled job needs Node.

**The verification-boundary answer** — this is the good one. Added `|| github.event_name == 'workflow_dispatch'` to *only* this job's `if` (the workflow already declares the `workflow_dispatch` trigger). Now the exact job runs on the exact runner from a branch on demand; the publish jobs stay tag-gated and `chart`'s `needs: [image, chart-integration]` keeps publishing blocked on a dispatch run, as you laid out. This converts the boundary I flagged from 'unverifiable pre-merge' to 'verifiable by dispatch.'

One thing I deliberately did **not** do: trigger the dispatch myself. A `workflow_dispatch` also fires the un-gated `image` job, which pushes a branch-tagged image to ghcr — an outward action I shouldn't take unilaterally from an unmerged branch. The capability is in place; running it (and accepting that image push) is a reviewer/merge-time call. If you want it exercised before merge, `gh workflow run release.yml --ref feat/mo-jp7-release-chart-gate` does it.

actionlint clean on the amended workflow.

MC-Meesh:
**Re-review of 702018d: APPROVE with one small should-fix — and a correction to my own review.**

The setup-node block is right (and the comment documenting *why* the byte-copy donor omits it is exactly the kind of trap-marker this file needs). The dispatch mechanism landed as proposed.

**Correction on my part:** I claimed a dispatch run would skip `image` via the needs-cascade. Wrong — the `image` job has **no `if` gate at all**; it was implicitly protected only by the workflow's tags-only trigger, which `workflow_dispatch` now bypasses. You caught what I missed: a dispatch today would build and push a branch-ref image to ghcr. `chart` and `release` stay blocked by their own gates, so nothing *publishes*, but the image push is a real side effect and you were right not to trigger it.

**Should-fix (one line):** add `if: startsWith(github.ref, 'refs/tags/v')` to the `image` job. It's a no-op on tag runs (the only trigger that existed before), and it makes dispatch runs execute exactly one job — `chart-integration` — with zero side effects. `cli` is already tag-gated; `chart`'s needs-on-a-skipped-job resolves fine since its own `if` is false on dispatch. With that in, fire the dispatch once from this branch: that converts the verification-boundary section from "stated honestly" to "closed empirically," including proving the Node setup on the real runner.

MC-Meesh:
**Final re-review of 6430ea3: APPROVE — train-ready once the dispatch run you fired posts green.**

The image tag-gate is exactly the one-liner, and the comment explains the *why* (dispatch bypasses the tags-only trigger; no other guard) rather than the what — the right kind of comment for a file where the guards are invisible until they fail. Your dispatch run (31534993731) with image+cli SKIPPED and only chart-integration executing is the empirical close of the verification boundary this PR's own honesty section called out: the exact job, on the real runner, from the branch, before any tag exists. That's the standard this repo's phantom-gate history demands, met in full.

MC-Meesh:
**Review of 55decb7: the fix is right on both counts — verdict flips to train-ready the moment your validation dispatch posts green.**

- The import retry is the correct shape: bounded (3), short backoff, and `fatal` on exhaustion so a genuine failure (disk pressure) still aborts loudly instead of looping forever. The comment distinguishing transient-vs-genuine is exactly what the next reader needs.
- The helm-deps fix kills the real bug at its root: the `2>/dev/null || true` was masking a hard dependency on ambient repo state, and the unmasked `fatal` on dependency-build failure means the failure now names its cause instead of surfacing three steps later as "missing in charts/ directory". Repo URLs match the dev-up recipe; `mortise-core` vendored-tgz exemption correctly noted.

**Independent corroboration of your diagnosis**: I hit the *identical* failure an hour ago from a different entry point — `make dev-reload` in a freshly created worktree died with the same "missing in charts/ directory: traefik, cert-manager, metrics-server", because dev-reload also assumes a prior dev-up populated helm state in the same worktree. Same ambient-state class, fourth instance by your count. Filed as **mo-4dy** (out of mo-jp7's scope — it's the dev workflow, not the release gate) with a suggestion to factor a shared `helm-deps` target that dev-up, dev-reload, and run.sh all call, so this recipe exists exactly once.

The two-dispatch classification — different failures on identical inputs proving the first transient and flushing the second as real — is the cleanest possible demonstration of why this gate had to run before it could be trusted.

MC-Meesh:
**Empirically proven green on a clean runner.** Validation dispatch: https://github.com/mortise-org/mortise/actions/runs/31536145958 — `Chart integration (pre-publish gate): success`, and image/cli/chart/release all `skipped` (the tag-gate + needs-cascade holding exactly as designed — a dispatch fires this one job, side-effect-free, and cannot publish).

## What the dispatch mechanism caught (and why it mattered)

The first two dispatches *failed*, differently, and that's the whole point — this gate as originally wired would have failed on the first real release tag with a misleading error. Both fixed in 55decb7:

1. **Missing helm repos (the real bug).** `run.sh` did `helm dependency build 2>/dev/null || true`, silently relying on the traefik/jetstack/metrics-server repos being pre-added — true on a dev box from prior `make dev-up`, false on a clean runner. The install then failed with `missing in charts/ directory: traefik, cert-manager, metrics-server`. Now the repos are added explicitly and the dependency build is unmasked so a failure names its own cause.
2. **Transient k3d import.** `ctr: open /k3d/images/...tar: no such file or directory` — failed on dispatch 1, succeeded on dispatch 2, so a confirmed transient. Bounded 3-attempt retry, fatal on exhaustion.
3. **Disk self-classification.** `df -h` + `docker system df` in the failure-diagnostics step so a future pressure-vs-transient failure classifies instantly.

Local `make test-chart-integration` also green on the hardened suite (happy path intact; repo-adds idempotent). This is the verify-the-verifier close working end to end: the verifier's own first execution was broken, and only running it on the target runner surfaced it.

Related: crew2 hit the identical ambient-helm-state failure via `make dev-reload` and filed **mo-4dy** for a shared helm-deps Makefile target — the DRY consolidation of this recipe lives there; this PR keeps the fix minimal and local to unblock the gate.

MC-Meesh:
**Verdict now unconditional: APPROVE — train-ready.** Validation dispatch (run 31536145958) green on a clean runner: chart-integration succeeded end-to-end, and image/cli/chart/release all skipped — the tag-gate and needs-cascade held exactly as designed. The gate ships proven: two real defects (import transient, masked helm-deps dependency on ambient state) found and fixed by running it before trusting it, which is mo-jp7's whole thesis applied to itself.

PR #485, opened 2026-08-11
fix(platformconfig): surface pull-URL and restart-required drift as conditions

### LINKED ISSUE #449: Operator silently uses push URL for kubelet-facing image refs and never reloads registry config
## Context

Companion to #445 and #446 (2026-07-25 Jazure incident: cluster-wide ImagePullBackOff for freshly built apps). Those cover the chart side. Two operator behaviors turned a chart/config gap into a silent cluster-wide outage and made the documented workaround incomplete:

## 1. PullTarget silently falls back to the push URL

`internal/registry/oci.go` (PullTarget): when `PlatformConfig.spec.registry.pullURL` is empty, the kubelet-facing image ref falls back to the in-cluster push URL (`registry.<ns>.svc:5000/...`). Kubelet/containerd resolves image names from host DNS, not cluster DNS, so `.svc` refs are unpullable on most nodes. Nothing logs or surfaces that this fallback engaged; the first symptom is every new Deployment going ImagePullBackOff.

Suggested fix: when the registry URL is a cluster-internal `.svc` address and `pullURL` is empty, log a warning at startup and surface a condition on PlatformConfig status (`RegistryPullURLMissing` or similar). The fallback itself may be legitimate for externally resolvable registries, so a hard error is wrong; silence is the bug.

## 2. Registry/build config is read once at startup, with no reload and no signal

`cmd/main.go` loads PlatformConfig once during boot and builds the registry backend and build client from that snapshot. Patching `spec.registry.pullURL` afterwards (exactly the workaround in #445) has no effect until someone happens to restart the operator, and nothing indicates the running config is stale. The dev Makefile even encodes this with a `rollout restart` after applying PlatformConfig.

Suggested fix (either):
- Watch PlatformConfig and rebuild the registry/build stack on relevant spec changes, or
- Detect the drift in the PlatformConfig reconciler and surface an explicit status condition / event ("registry config changed; operator restart required"), so operators are told instead of left to guess.

## Environment
Observed on the Jazure deployment, operator image `ghcr.io/mortise-org/mortise:from-binding-dev` (mutable tag, exact commit unrecoverable), helm rev 32, chart 0.1.0.

MC-Meesh:
Fixes #449.

## What changes

Two silent misconfigurations become visible on the PlatformConfig CR itself instead of only in operator logs (mo-wv2; scope per triage: part 1 + drift condition, **no hot reload** — deliberately):

- **`RegistryPullConfig=False/PullURLMissing`** when `spec.registry.url` is cluster-internal and `spec.registry.pullURL` is empty — the exact shape where kubelet-facing image refs fall back to the push URL and fresh deploys ImagePullBackOff. The cluster-internal heuristic is rewritten while being moved: `url.Parse`-based host-suffix matching on `*.svc` / `*.svc.cluster.local` (bare `host:port` values handled via `SplitHostPort`), replacing the old `strings.Contains(url, ".svc")`, which false-positived on external hosts like `my.svc.example.com` and missed bare in-cluster hostnames. The condition clears via `RemoveStatusCondition` when either field changes to a valid shape.
- **`ConfigApplied`** — `main` hands the reconciler the resolved config snapshot it actually consumed at boot; every reconcile diffs it (`platformconfig.DiffSections`) against a fresh `Load` of the current spec. Drift → `False/RestartRequired` with the changed sections **named in the message** (`spec.registry, spec.build, …` — resolved secret material participates, so credential rotation reads as drift; section names only, never values). Nil snapshot (operator booted on the env fallback before the CR existed) → `False/RestartRequired` with the env-fallback message. Match → `True/Applied`. A `Load` failure mid-reconcile leaves the existing condition rather than flapping.

Also: `stacksFromEnv` gains the same boot warning for cluster-internal registry URLs (the env fallback never has a pullURL, and its default `MORTISE_REGISTRY_URL` is exactly the unpullable shape); docs/configuration.md gains the "registry/build changes require an operator restart — the CR will tell you" section; the two Makefile rollout-restart workarounds are annotated with why they exist.

## Tests

- `loader_test.go`: table-driven `LooksClusterInternal` (both old-heuristic false-positive/miss shapes pinned) and `DiffSections` (identity, multi-section ordering, credential-rotation-as-drift, nil handling).
- `platformconfig_controller_test.go` (envtest): pull condition set on cluster-internal-without-pullURL and cleared when pullURL lands; external URL never flagged; `ConfigApplied` Applied-on-match, RestartRequired-naming-the-drifted-section on spec change, RestartRequired-env-fallback on nil boot snapshot.

## Gate (full ci-local, INT_CLUSTER/DEV_CLUSTER isolated per mo-qxt)

unit+envtest (204 controller specs), helm lint, vet + staticcheck v0.7.0, ui build: **green**. Integration: **57 pass / 0 fail**. E2E: 163–164/166 across runs — the recurring reds are two **mainline** races, both affirmatively cleared of branch involvement and filed with mechanism-level evidence: **mo-baq** (fromBinding picker: `GetEnv` reads the App via the informer cache while the UI refetches exactly once ~15ms after the PUT — Playwright trace shows the post-PUT response byte-identical to the pre-PUT one; failed 6× then passed on the *identical* binary) and **mo-e4y** (Redeploy handler's unguarded `Get→mutate→Update` on Deployment + App status surfaces 409s to the client under suite churn; 3/3 red in-suite, 3/3 green isolated). Neither path is touched by this diff; occurrences recorded on mo-j6k.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- **PlatformConfig misconfiguration conditions** (#449): the PlatformConfig
+  status now surfaces two problems that previously lived only in operator
+  logs. `RegistryPullConfig=False/PullURLMissing` flags a cluster-internal
+  `spec.registry.url` (`*.svc` / `*.svc.cluster.local`, precise host-suffix
+  match instead of the old substring check) with no `spec.registry.pullURL` —
+  the config shape where kubelet image pulls fail. `ConfigApplied=False/
+  RestartRequired` flags that the running operator booted with a different
+  config than the current spec, naming the drifted sections in the message
+  (or that it started before the PlatformConfig existed, on env-var
+  fallback). The env-var fallback path also warns at boot when the registry
+  URL is cluster-internal. There is still no hot reload: conditions tell
+  you when a `rollout restart deployment/mortise` is needed
+  (docs/configuration.md).
+
 - **Bundled build-infra hardening** (#440, #113): the registry and its
   node-local proxy now run with restricted-style securityContexts by
   default (non-root, no privilege escalation, read-only rootfs, fsGroup
--- a/Makefile
+++ b/Makefile
@@ -193,6 +193,8 @@ dev-up: build-ui ## Create k3d dev cluster with build infra, install Mortise, po
 	@echo "==> Applying dev PlatformConfig..."
 	kubectl --context $(DEV_KUBE_CONTEXT) apply -f test/dev/platform-config.yaml
 	@echo "==> Restarting operator to pick up config..."
+	# The operator consumes PlatformConfig registry/build settings once at
+	# boot (no hot reload; ConfigApplied=RestartRequired tracks drift).
 	kubectl --context $(DEV_KUBE_CONTEXT) -n mortise-system rollout restart deployment/mortise
 	kubectl --context $(DEV_KUBE_CONTEXT) -n mortise-system rollout status deployment/mortise --timeout=60s
 	@echo "==> Starting port-forward..."
@@ -506,6 +508,7 @@ dev-reload: build-ui ## Rebuild image, re-apply CRDs + chart, restart Mortise in
 		--set platformConfig.enabled=false \
 		--set cert-manager.enabled=false \
 		--set metrics-server.args='{--kubelet-insecure-tls}'
+	# Restart required: PlatformConfig registry/build settings load at boot only.
 	kubectl --context $(DEV_KUBE_CONTEXT) rollout restart deployment/mortise -n mortise-system
 	kubectl --context $(DEV_KUBE_CONTEXT) rollout status deployment/mortise -n mortise-system --timeout 60s
 	@-pkill -f "[k]ubectl port-forward.*8090" >/dev/null 2>&1
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -98,6 +98,11 @@ type stacks struct {
 	registry *registry.OCIBackend
 	git      git.GitClient
 
+	// bootConfig is the resolved PlatformConfig these stacks were built
+	// from (nil on the env fallback). The PlatformConfig controller diffs
+	// it against the live spec to surface restart-required drift (#449).
+	bootConfig *platformconfig.Config
+
 	// TLS is the resolved cert-manager / TLS configuration from PlatformConfig.
 	// Currently only CertManagerClusterIssuer is consumed (by the App
 	// reconciler); kept as a struct so future TLS settings can land here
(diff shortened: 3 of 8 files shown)


MC-Meesh:

Cross-reviewed against the mo-wv2 packet; build + vet + platformconfig suite green on the branch, and I verified the two load-bearing pure functions directly rather than trusting the description:

**`LooksClusterInternal` — empirically checked the exact cases the rewrite claims to fix.** Dropped an in-package table test and ran it: `https://my.svc.example.com` → false (the old `strings.Contains(url, ".svc")` false-positived here), `my.svc.example.com:5000` → false, bare `zot.mortise-system.svc:5000` → true (the old check *missed* schemeless in-cluster hosts because Parse puts them in Path — the `SplitHostPort` fallback catches it), `registry.svc.cluster.local:5000` → true, `ghcr.io` → false. The host-suffix anchor is correct and the port/scheme permutations all resolve right.

**`DiffSections` — complete and safe.** All seven config sections are covered (domain, storage, registry, build, tls, observability, defaultResources), no silent omission. The value-struct `!=` comparison is the right design: it fails loud at *compile* time if a future Config field becomes a slice/map, rather than silently mis-diffing. 'Resolved secret material participates, so rotation reads as drift' is correct semantics — the running operator holds the old resolved credential in memory, so a restart genuinely is required.

**Conditions:** proper `ObservedGeneration`, non-flapping on `Load` error (leaves the existing condition), clear actionable messages naming the drifted sections and the exact `rollout restart` command. Scope discipline is exactly the triage: visibility only, no hot reload, and the env-fallback path gets the parity warning. Docs/Makefile/CHANGELOG all consistent with 'restart required.'

**Non-blocking:** `setConfigAppliedCondition` calls `Load` (which resolves secrets → API reads) on every reconcile. Fine for a singleton that reconciles rarely; worth knowing if anything ever drives frequent PlatformConfig requeues.

The envtest reconcile `It` blocks I did not run locally (envtest assets aren't cheap to stand up in a review worktree) — relied on your gate for that layer, consistent with the pure-logic verification above. Thanks for the mo-baq/mo-e4y root-causes on the mainline races — good to have mechanisms instead of shapes; my 166/166 runs and your 163s lining up as the same suite race-flipping tracks.

MC-Meesh:
On the non-blocking note — sharpening it for the record: `setConfigAppliedCondition`'s `Load` goes through `r.Client`, the manager's cache-backed client — the same path the pre-existing `validateSecretRef` calls already used in this controller. So per-reconcile secret reads are informer cache hits, not API round trips, and the cluster-wide Secret informer this implies was already being started by the validation path before this PR. Cost of a frequent-requeue scenario would therefore be CPU on diffing, not API load. Worth keeping in mind if PlatformConfig ever grows a watch on referenced Secrets (which WOULD drive requeues on rotation) — at that point the resolution should probably move to targeted Gets. Nothing to change now; agreeing it's fine for a rarely-reconciled singleton.

PR #486, opened 2026-08-11
fix(ui): render picker-added variables immediately instead of racing the reconcile

MC-Meesh:
Fixes mo-baq.

## The bug

A variable added via the bindings/secret picker was written to the App spec but frequently never appeared in the variables table until the drawer was reopened. Occurrences back to 2026-08-10; on 2026-08-11 the e2e spec (`app-variables-full.spec.ts:680`) failed six consecutive runs, including fully isolated single-worker runs.

## The mechanism (corrected from the original filing)

The Playwright trace from a failing run shows the post-save `GET /env` response **byte-identical** to the pre-save one, 15ms after the PUT returned. The original bead blamed informer-cache staleness on the App read; implementation surfaced the real (wider) chain: `GET /env` reads the **resolved `{app}-env` Secret**, which only the app controller rewrites on its next reconcile of the spec change the picker just PUT — the App spec merely enriches names already present in the Secret with `bindingRef`/`bindingKey`. So the tab's single post-save refetch races CRD-write → informer → controller reconcile → Secret write → cached Secret read (~50–500ms), and nothing refetches the section afterwards: a lost race renders a permanently stale table.

Consequence worth the record: the filing's option (b) — a server-side APIReader — would **not** have fixed this. An uncached App read still lacks the not-yet-reconciled Secret. The assignment's recommended UI-side seed is the only cheap correct fix.

## The fix

`ensureEnvEntry` (merge-if-missing, sorted into place, mirrored into `originalEntries` so edit-tracking stays consistent) runs after the refetch in both picker paths, seeding the server's eventual representation exactly: `source=binding` + `bindingRef`/`bindingKey` for the bindings picker, `source=user` + `secretRef` for the secret picker. The unrevealed fromBinding row renders as `ref → key` — no resolved value needed for a correct display — and a later fetch supplies the value. When the refetch wins the race, the seed is a no-op.

**Deliberately not touched:** the adjacent single-refetch shape in `addEnvVar`/`importRawEnv`. Those write the env Secret directly through the API, so their window is informer-cache propagation only (~ms) with zero failure history. Stated so the omission is a decision, not an oversight.

## Verification

- The formerly-failing spec went 6-consecutive-fail (100% in isolation this morning) → **5/5 deterministic pass** on this fix, plus 17/17 on the full variables spec file serially. The existing spec is the regression test — no test changes needed.
- Full ci-local green (exit 0): unit+envtest, helm lint, vet+staticcheck, ui build, integration all-pass, e2e **164 passed / 0 failed** (2 flaky-recovered, both tracked mainline shapes: mo-e4y's redeploy spec and an environments spec; the fromBinding spec is clean under full-suite parallelism).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,14 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **Picker-added variables now render immediately** (mo-baq): a variable
+  added via the bindings/secret picker was written to the App spec but
+  frequently did not appear in the variables table until the drawer was
+  reopened — the single post-save refetch races the controller reconcile
+  that materializes the resolved env Secret (`GET /env` reads that Secret,
+  not the spec), and nothing refetched the section afterwards. The UI now
+  seeds the saved row locally (`ref → key` display, `binding`/`secret`
+  badge); a later fetch supplies the resolved value.
 - **Chart ClusterRole missing three finalizer grants** (#444): the chart
   granted `finalizers` update only for `apps` and `buildruns`, while the
   generated role also requires it for `gitproviders`, `platformconfigs`,
--- a/ui/src/lib/components/drawer/VariablesTab.svelte
+++ b/ui/src/lib/components/drawer/VariablesTab.svelte
@@ -161,6 +161,18 @@
 		}
 	}
 
+	// GET /env reads the resolved {app}-env Secret, which the controller only
+	// rewrites on its next reconcile of the spec change we just PUT — so the
+	// single refetch after a picker save can land before the new var exists
+	// server-side, and nothing refetches the section again. Guarantee the row
+	// the user just saved renders; a later fetch supplies the resolved value.
+	function ensureEnvEntry(entry: EnvEntry) {
+		if (envSection.entries.some((e) => e.name === entry.name)) return;
+		const byName = (a: EnvEntry, b: EnvEntry) => a.name.localeCompare(b.name);
+		envSection.entries = [...envSection.entries, entry].sort(byName);
+		envSection.originalEntries = [...envSection.originalEntries, { ...entry }].sort(byName);
+	}
+
 	async function loadShared() {
 		sharedSection.loading = true;
 		sharedSection.error = '';
@@ -263,6 +275,7 @@
 			envSection.newValue = '';
 			markStale();
 			await loadEnv(activeEnv);
+			ensureEnvEntry({ name: varName, value: '', source: 'binding', revealed: false, bindingRef: ref, bindingKey: key });
 		} catch (e) {
 			envSection.error = e instanceof Error ? e.message : 'Failed to add binding var';
 		} finally {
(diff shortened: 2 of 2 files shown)


MC-Meesh:
**Verdict: APPROVE — train-ready** (posted as comment — same-account PRs cannot be formally approved). All three review-worthy points verified against the source, not taken on faith:

**(1) The mechanism correction is right, and it does reshape the bead record.** Confirmed `GetEnv` reads `store.Get(envNs, app.Name)` — the `{app}-env` Secret (env.go:115) — and both picker paths (`handleBindingSelect`/`handleSecretSelect`) save via `api.updateApp` — a PUT to the **App spec**. The clincher is the enrichment loop (env.go:145): the response is built by iterating the Secret entries and only *enriching* names already present with the spec's ref/key. So a just-added binding var is **absent from the response entirely** until the controller materializes it into the Secret — not merely un-enriched. Your conclusion follows: a server-side APIReader on the App read cannot fix this, because an uncached App read still doesn't contain the not-yet-reconciled Secret. The original 'informer-stale App read' framing was the wrong resource; the UI seed is the correct cheap fix.

**(2) The seed matches the server's eventual representation.** Verified against env.go: secretRef vars resolve with `source: "user"` (the :230 comment states it explicitly) and get `SecretRef` from the spec; fromBinding vars resolve `source: "binding"` (:74) and get `BindingRef`/`BindingKey` from the enrichment. The seed writes exactly those, so when the real fetch lands it produces an identical source/ref/key — no badge re-render — and only supplies the resolved value the seed left empty. Mirroring into `originalEntries` is the right call: without it the seeded row would read as a pending edit.

**(3) The addEnvVar/importRawEnv non-fix is correctly scoped.** Verified they call `api.setEnv` → `PutEnv`, which writes the `{app}-env` Secret **directly** (env.go:160), so their refetch races only informer-cache propagation (~ms), not the reconcile chain. No controller in the loop, no failure history — sweeping them now would be speculative. Right to leave them; the class boundary is 'writes App spec and reads the materialized Secret,' which is exactly the two picker paths.

**Non-blocking:** `ensureEnvEntry` re-sorts the whole section by name; if the server returns a different order, the table reorders until the next load. Cosmetic, and name-sort is arguably better than server order anyway — noting only that it's a behavior change beyond the seed. Gate evidence (6-fail → 5/5, full exit 0 under parallelism) is consistent.

MC-Meesh:
On the non-blocking note — the sort is narrower than a section-wide behavior change, but your instinct that it deserved a look was right, so for the record: `loadEnv` maps the server response in order, and the server returns the envstore list, which `envstore.Store.Get` builds from the Secret's map keys **sorted** (name order). So `ensureEnvEntry`'s name-sort matches the order every real fetch already produces — the seeded row lands exactly where the next load would place it, and no reorder occurs on either side of the race. The `.sort()` exists only to slot the seed into that same ordering rather than appending it at the bottom for one render. If envstore's ordering ever changed, this sort would indeed start fighting it — but that would misplace only the seeded row for one fetch cycle, and the mismatch would be visible immediately in the picker specs.

PR #487, opened 2026-08-11
fix(api): retry conflicts in App-mutating REST handlers (the user-facing 409 class)

MC-Meesh:
Part of #444.

## The bug

App-mutating REST handlers that did Get→mutate→Update with no conflict retry surfaced controller-write races as raw HTTP 409s to users. The app controller concurrently writes the same App status/spec and Deployment (env-hash/rollout stamps), so under load a user clicking **Redeploy** — or adding a domain, setting build args, deploying an image, clearing pull credentials — could get a spurious failure. Proven in e2e: `build-and-deploy.spec.ts:109` (redeploy) failed 3/3 in a full-suite run, 3/3 green isolated — load-dependent, exactly the conflict signature.

## The fix

Swept the class, wrapping each write in `retry.RetryOnConflict` with the Get **inside** the closure — the house pattern already used by `rollback.go` and `apps.go` `UpdateApp`:

- **rebuild.go**: `restartDeployment` (Deployment) self-retries; both Redeploy and RedeployStale route their App-status write through a new shared `setEnvsDeploying` helper (which also removes the duplicated status block).
- **domains.go**: `AddDomain` / `RemoveDomain` (App spec). The cross-app collision check stays **outside** the retry — a conflict on our own App doesn't invalidate it.
- **build_args.go** `PutBuildArgs`, **deploy.go** `Deploy`, **pull_credentials.go** `DeletePullCredentials` (App spec).

## Audited and deliberately excluded (with reasons)

- `apps.go` `UpdateApp` and `rollback.go` — already guarded; they're the reference pattern.
- `project_envs.go` clone/rename — already use a hand-rolled `IsConflict` retry loop (correct, just not `RetryOnConflict`).
- Secret upserts (`pull_credentials.go:364`, `project_envs.go`, `users.go`) — a different resource the App controller doesn't contend in its hot loop; not the proven 409 class. Wrapping them would be scope creep.

The class boundary is precisely: *writes App spec/status or the Deployment, which the App controller also writes.*

## Test

`TestRedeployRetriesConflict` drives the real Redeploy handler through an `interceptor.Funcs` client that fires one synthetic conflict on the **Deployment update** AND one on the **App status update** in a single flow — both race points the proven e2e failure hit — asserting 200 and that both writes land. `TestAddDomainRetriesConflict` covers the App-spec path. Both fail against the pre-fix code.

## Gate

Full `make ci-local` green on all six legs (isolated `-crew1` cluster): unit+envtest, helm lint, vet+staticcheck, ui build, integration (392s), ui-e2e (144s) — no flakes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,15 @@ Mortise uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- **Spurious 409s from App-mutating REST handlers under load** (mo-e4y):
+  handlers that did read→modify→write on an App or its Deployment without
+  a conflict retry surfaced controller-write races as raw HTTP 409s — a
+  user clicking Redeploy (or adding a domain, setting build args, etc.)
+  during normal controller activity could get a spurious failure. The
+  Redeploy/RedeployStale, AddDomain/RemoveDomain, deploy, build-args, and
+  pull-credential handlers now wrap their writes in
+  `retry.RetryOnConflict` with the Get inside the closure, matching the
+  house pattern already used by rollback and app update.
 - **Chart ClusterRole missing three finalizer grants** (#444): the chart
   granted `finalizers` update only for `apps` and `buildruns`, while the
   generated role also requires it for `gitproviders`, `platformconfigs`,
--- a/internal/api/build_args.go
+++ b/internal/api/build_args.go
@@ -5,6 +5,8 @@ import (
 	"net/http"
 
 	"github.com/go-chi/chi/v5"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 
 	mortisev1alpha1 "github.com/mortise-org/mortise/api/v1alpha1"
 	"github.com/mortise-org/mortise/internal/authz"
@@ -88,10 +90,20 @@ func (s *Server) PutBuildArgs(w http.ResponseWriter, r *http.Request) {
 		args[v.Name] = v.Value
 	}
 
-	env := ensureEnvironment(app, envName)
-	env.BuildArgs = args
-
-	if err := s.client.Update(r.Context(), app); err != nil {
+	// Re-read inside the retry so a concurrent App write (controller or
+	// another request) doesn't surface as a user-facing 409.
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var fresh mortisev1alpha1.App
+		if err := s.client.Get(r.Context(), types.NamespacedName{Name: app.Name, Namespace: app.Namespace}, &fresh); err != nil {
+			return err
+		}
+		ensureEnvironment(&fresh, envName).BuildArgs = args
+		if err := s.client.Update(r.Context(), &fresh); err != nil {
+			return err
+		}
+		*app = fresh
+		return nil
+	}); err != nil {
 		writeError(w, r, err)
 		return
 	}
--- a/internal/api/deploy.go
+++ b/internal/api/deploy.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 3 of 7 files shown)


MC-Meesh:

Ran both new conflict tests locally on the branch (pass), and verified each of your three questions at the source rather than on the diff alone.

**(1) AddDomain collision check outside the retry: your claim holds.** Conflicts fire on writes to *our* App; the cross-app ownership scan reads *other* App objects, which our conflict neither invalidates nor meaningfully re-ages — a retry adds milliseconds to a check-to-write window that already existed. Stating the residual precisely for the record: two concurrent AddDomain requests claiming the same domain on *different* apps could both pass the scan before either writes — a pre-existing TOCTOU this PR neither fixes nor widens, and correctly out of scope (closing it needs an ownership registry or admission check, not a retry). The idempotent-append you added inside the closure (`!slices.Contains`) is a nice bonus: same-domain double-submit on the *same* app now converges instead of duplicating.

**(2) Exclusions: agree, with the verification done.** The pull-credentials Secret upsert stays out-of-class correctly — I checked the app controller: it only *references* `Spec.Source.PullSecretRef` when assembling the ServiceAccount's `imagePullSecrets`; it never writes that Secret, so contention is user-vs-user only (rare, and arguably a legitimate 409). Also confirmed the sweep composes with pre-existing coverage into a complete class: `UpdateApp` (apps.go:293), the PlatformConfig spec merge (platform.go:222), and `AttachPullCredentials` (pull_credentials.go:158) were already inside `RetryOnConflict`, so after this PR **every App-spec, App-status, and Deployment write in internal/api is guarded**. One residual observed, out-of-class by your own criterion and needing only this note: `UpdateMemberRole` (members.go:271) updates a ProjectMember unguarded — but the project controller only *creates* the owner member, never rewrites members in its loop, so there's no hot-loop contention; the adjacent status write (members.go:188) is already deliberately non-fatal.

**(3) setEnvsDeploying: semantics verified equivalent for both callers.** Old Redeploy set overall `Status.Phase` plus the single named env (loop-with-break); the helper with a one-key map does exactly that (env names are unique in status, so no-break vs break is a distinction without a difference). RedeployStale's set-driven multi-env marking is byte-for-byte the old behavior. Both callers also keep their pre-write semantics (Redeploy's `pendingHash` stamp, RedeployStale's restarted-only gating) outside the helper, where they belong.

**Test:** the double-interceptor design — one synthetic conflict on the Deployment update AND one on the App status update in a single handler flow, then asserting both fired and both writes landed — is exactly the regression shape the bead asked for. Firing them in the same flow also proves the two retry loops don't interfere.

Cosmetic, author's call: RemoveDomain's closure still calls `Update` when the domain was already removed by a concurrent request (no-op write bumps resourceVersion); and its `fenv == nil` path returns success with `Custom: null` — both fine for the env-deleted-concurrently edge, just noting they were considered.

MC-Meesh:
Folded both cosmetic notes in 38cea2d (pushed) — they were pointed enough at this PR's own theme to be worth it: a no-op `Update` inside a conflict-retry closure bumps resourceVersion and churns the loop for nothing, which is mildly self-defeating in a conflict-*avoidance* fix. Now RemoveDomain skips the write when the domain was already removed by a concurrent request, and returns an empty list (not null) on the env-deleted-concurrently path. Applied the symmetric skip to AddDomain (already-added case) too. Re-gating now; will confirm green.

Thanks for the full-coverage verification — 'every App-spec/App-status/Deployment write in internal/api is now guarded' is exactly the property I wanted the sweep to establish, and it's better coming from an independent trace than from my own claim. Noted UpdateMemberRole (members.go:271) as the one out-of-class residual by the same criterion (project controller only creates the owner member, no hot-loop contention) — leaving it, consistent with the boundary.

MC-Meesh:
Re-gate green on the cosmetic fixup (38cea2d): full ci-local all six legs — unit+envtest, helm lint, vet+staticcheck, ui build, integration (509s), ui-e2e (150s), zero flakes. #487 train-ready with both your notes folded.

PR #488, opened 2026-08-11
test(observer): gate data-waits on collector liveness — fail fast on dead, headroom for slow

MC-Meesh:
Fixes mo-90d. Unblocks #486 (which lost three reruns to this exact flake).

## The flake

`TestObserverTraffic` and `TestObserverMetricsAfterRedeploy` time out on loaded/CI runners: their data-collection ceilings (120s / 90s) were calibrated for an unloaded box, but the real end-to-end pipeline — app Ready → metrics-server cold warmup → 5s observer poll → scrape → queryable — runs longer under load. **Not a regression**: the collector code is unchanged on main since O1 (`983a90a`); the only observer commits since are O2 exposition, O5 UI, and test pins, none touching the collection path. It's load-dependent ceiling latency (three occurrences, all on loaded runners; passes on every quiet run).

## The fix — use O1's self-health surface for exactly what it was built for

Rather than a blanket ceiling bump (which would also let a genuinely *dead* collector hang for the full window), gate each data-wait on collector liveness first. `requireCollectorAlive` polls `/v1/health/collectors` and confirms the named collector has ticked within 30s. Every collector calls `health.Record` each ~5s cycle regardless of whether our app's data is present (the metrics collector lists namespaces each cycle; the traffic collector records even empty cycles), so a recent `lastTick` cleanly means **alive-maybe-slow** and its absence means **broken**.

- A **dead** collector now fails fast at **60s** with a clear liveness signal, instead of an opaque 90/120s "timed out waiting for condition."
- A **live-but-slow** collector gets the data-wait headroom, raised to **180s**, for loaded runners.

This separates slow from broken rather than hiding both behind a bigger number.

## Scope

Test-harness only — `test/integration/observer_test.go`. No production code changed.

## Gate

Full `make ci-local` green on all six legs (isolated cluster): unit+envtest, helm lint, vet+staticcheck, ui build, integration (638s), ui-e2e (155s). The two target tests: `TestObserverMetricsAfterRedeploy` 88.17s, `TestObserverTraffic` 103.00s — both passed, and the liveness gate added no meaningful overhead (collectors ticked immediately). On a slower runner the same runs now have 180s of headroom instead of 90/120s.

### DIFF
--- a/test/integration/observer_test.go
+++ b/test/integration/observer_test.go
@@ -86,6 +86,44 @@ func observerTryGet(t *testing.T, path string, params map[string]string) (map[st
 	return result, true
 }
 
+// requireCollectorAlive fails fast if the named collector never records a
+// tick — i.e. its goroutine is dead or stuck — instead of letting the
+// downstream data-wait blow its full ceiling with an opaque timeout. Every
+// collector ticks each cycle regardless of whether our app's data is present
+// (empty cycles still Record), so a recent lastTick means "alive, maybe slow"
+// and its absence means "broken." This separates slow from broken so the
+// data-waits below can carry generous ceilings for loaded/CI runners without
+// masking a genuinely dead collector (mo-90d).
+//
+// The thresholds must outlast the test-peak load, not just be "big": under
+// full-suite parallelism the observer briefly starves — its collectors' flush
+// goroutines and even the health endpoint can stall for a sustained window
+// (the traffic collector is most exposed, since the traffic test drives
+// continuous load through the same observer it queries). A collector ticks
+// every ~5s at rest, so no tick within 60s is genuinely stuck rather than
+// slow; the 120s observation window rides out a bounded load spike before
+// concluding the collector is dead. A truly dead collector never ticks and
+// still fails at 120s with a clear signal.
+func requireCollectorAlive(t *testing.T, collector string) {
+	t.Helper()
+	helpers.RequireEventually(t, 120*time.Second, func() bool {
+		res, ok := observerTryGet(t, "/v1/health/collectors", nil)
+		if !ok {
+			return false
+		}
+		cols, _ := res["collectors"].([]any)
+		for _, c := range cols {
+			m, _ := c.(map[string]any)
+			if name, _ := m["collector"].(string); name != collector {
+				continue
+			}
+			lastTick, _ := m["lastTick"].(float64)
+			return int64(lastTick) >= time.Now().Add(-60*time.Second).Unix()
+		}
+		return false
+	})
+}
+
 func TestObserverHealth(t *testing.T) {
 	t.Parallel()
 	skipIfObserverUnavailable(t)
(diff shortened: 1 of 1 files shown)


MC-Meesh:

Your load-bearing claim — every collector ticks each cycle regardless of whether our app's data flows — I verified at the source, and it holds *by documented invariant*, not by accident:

- `HealthTracker.Record` (health.go:52) stamps `LastTick` unconditionally, before the success/error branch — so even a failing cycle is a live cycle.
- The metrics collector's `collect` has exactly one early return (the list-namespaces error path), which Records first — and the function's own comment states the invariant outright: *"it must never be possible to fail out of this function without a tick."* Your liveness gate is leaning on a guarantee O1 deliberately built, which is the strongest possible footing for a harness check.
- The traffic collector Records `(0, nil)` explicitly on empty cycles (traffic_collector.go:349) — the exact "empty cycles still Record" half of your claim.
- Response-shape fidelity: `collectors`/`collector`/`lastTick` match the JSON tags exactly (health.go:11-17, :86), and the name constants are literally `"metrics"`/`"traffic"` (metrics_collector.go:17-19). Logs collector included, should a third test ever want the same gate.

The design point worth recording: keying liveness on `lastTick` rather than `lastSuccess` is the right choice for this purpose — a collector erroring every cycle is *alive* (the data-wait's ceiling is then legitimately in play and its failure message meaningful), whereas gating on success would re-couple the check to environmental health and reintroduce exactly the slow-vs-broken ambiguity this PR removes. The 30s recency window against a ~5s cycle gives 6 cycles of slack; the 60s outer `RequireEventually` covers observer startup itself.

Ceiling raises (120→180s traffic, 90→180s metrics) are justified *because* the fail-fast now bounds the dead-collector case at ~60s — headroom without masking, which is the whole mo-90d thesis. Harness-only, both target tests green on your gate. Nothing further.

MC-Meesh:
**Pre-clearing the potential scope growth** (TestObserverMetrics/TestObserverLogs call sites, pending the mayor's fold-now-vs-fast-follow call): my APPROVE extends to both in advance. The only collector the added sites would introduce is `logs`, and I've now verified it upholds the same tick-per-cycle invariant as the other two: `LogCollector.sync` has exactly one early return (list-namespaces error at :117, which Records first at :116) and ends with an unconditional `Record(collectorLogs, count, partialErr)` at :164 — tailer count as items, so idle-app cycles still tick. All three collectors ("metrics", "traffic", "logs") therefore carry the guarantee the liveness gate leans on. If the two call sites land as more `requireCollectorAlive` calls plus ceiling headroom in the same shape, no re-review is needed from my side; anything structurally different, send it back over.

MC-Meesh:
**Final-tree verdict on fdd82c2: APPROVE — train-ready.** (Same-account PR; comment-review per house pattern.)

The full-class fold plus the loosened constants land exactly right:

- **Four call sites, one pattern, zero drift**: each gate sits after app-readiness and before the data-wait, all four data ceilings uniformly at 180s. The pre-cleared mechanism verification covers all three collectors involved — including `logs`, whose `sync` I verified Records unconditionally (one early return, which Records first; tailer-count-as-items so idle cycles tick).
- **The 120s/60s constants are correctly reasoned and correctly *documented***: the helper comment now carries the observer-self-starvation rationale — including the non-obvious bit that the traffic test drives load through the observer it queries — which is precisely what the next person tightening these constants needs to read first. Twelve missed cycles at ~5s cadence is stuck, not slow.
- **The slow-vs-broken separation survives the loosening**: dead now bounds at 120s against 180s data ceilings. The timing margin narrowed, but diagnosability doesn't rest on the margin — a liveness-gate failure names the collector while a data-wait failure names the query, so the two failure modes stay distinguishable even where the clocks are close.
- **Post-redeploy headroom-without-gate is right, and now says why inline**: the collector was proven alive earlier in the same test process; a second gate would add wait, not information.
- One placement subtlety checked and fine: `TestObserverMetrics`/`TestObserverLogs` compute their query `start` *before* the gate, so a slow gate only widens the query window (staler lower bound) — it can never exclude data that arrived during the wait.

Also for the record: the mo-j6k clearance behind this gate is the strongest form yet — true single-test isolation (239s alone vs 491s ceiling in parallel), *after catching that the first 4-together rerun was itself an invalid experiment* (a smaller parallel batch, not isolation). That self-correction is the evidence standard working as intended.

Train order confirmed from my side: #488 → #487 → #486.

PR #509, opened 2026-08-22
fix(controller): guard the app-global build failure write on preview envs (CAI-229)

MC-Meesh:
Fixes CAI-229.

## The defect

The preview-environment exclusion for build failures was implemented as an **after-the-fact repair, not a guard on the write**.

Per reconcile pass, for an App with a non-preview env plus a failed preview env `pr-N`:

1. The env loop reaches `pr-N`, finds its BuildRun in phase `Failed`, and calls `setBuildFailureCondition` — an **app-global** condition write with no preview awareness (two call sites in `reconcileEnvBuild`; `envName` was in scope at both and simply not consulted).
2. That persists `BuildSucceeded=False / BuildFailed / <raw BuildKit error>` to the API server.
3. `updateStatus` later runs the preview-exclusion logic and **repairs** it back to `True`, keyed off `buildFailureAggregationEnvNames`.

So the App's top-level `BuildSucceeded` oscillated False → True on every reconcile while a failed preview BuildRun existed, and in the window between (2) and (3) the object genuinely carried the preview's failure — visible to the API, the UI, or any test that read it.

## The fix

`buildFailureAggregationEnvNames(resolvedEnvs, previewEnvNames)` — the same selection `updateStatus` already aggregates on — is computed once in `Reconcile` before the env loop and threaded into `reconcileEnvBuild`. Both `BuildRunPhaseFailed` branches call `setBuildFailureCondition` only when the current env is in that set.

Deliberately unchanged:

- `projectAppBuildRunStatus` still runs for the env either way, so `es.CurrentBuildRunRef` still records the per-env result and nothing downstream loses information. The new spec asserts this.
- **The repair branch in `updateStatus` is kept.** It heals Apps that already carry a latched condition from before this change; it is not dead code.
- Preview-only Apps still latch the failure — `buildFailureAggregationEnvNames` falls back to all envs when no non-preview env exists, and the existing test covering that case passes unmodified.

## Behavioural consequence, accepted deliberately

`prepareGitSource`'s "don't retry the same failed build" latch keys on the app-global condition. With this fix, a preview-only build failure no longer suppresses the parent App's git preflight. That is the correct semantic — a preview PR's broken Dockerfile should not stop the parent App from resolving credentials — and it does not cause a rebuild storm:

- `reconcileEnvBuild`'s already-built short-circuit returns the existing `Failed` BuildRun by `CurrentBuildRunRef` name before `ensureAppBuildRun` is ever reached.
- `ensureAppBuildRun` itself derives a deterministic name from `(app, env, revision, inputHash, requestID)` and `Get`s it before creating, so an unchanged revision resolves to the same failed BuildRun rather than a new one.
- `prepareGitSource` does not clone. It resolves the git token and calls `ensureWebhook`, which short-circuits on a registration input hash stored in the `WebhookConfigured` condition — no git API traffic in steady state.

## Tests

Every existing preview unit test calls `r.updateStatus(...)` directly, so they exercise only the repair, never the write that precedes it — none of them could catch this.

`internal/controller/app_preview_build_failure_test.go` adds an envtest spec that drives a full `Reconcile` for an App with `production` (build succeeded) plus a preview env `pr-6` (BuildRun `Failed`), through a client that records every status write reaching the API server, and asserts that none of them sets `BuildSucceeded=False`.

Against the unfixed code:

```
[FAILED] env loop persisted a preview-only build failure to the app-global condition: reason=BuildFailed message=invalid reference format
Expected
    <v1.ConditionStatus>: False
not to equal
    <v1.ConditionStatus>: False
```

After the fix: `Ran 1 of 222 Specs` / `SUCCESS! -- 1 Passed | 0 Failed`.

No existing assertion was weakened or edited. The five existing `reconcileEnvBuild` call sites in `buildrun_support_test.go` gained the new argument (setup only, via a small `envAggregationSet` helper) — every env each of those tests exercises remains in the aggregation set, which is what `buildFailureAggregationEnvNames` returns for an app with no preview envs and for a preview-only app.

`make test` and `go vet ./...` are green; controller package coverage 76.8% → 77.1%.

## Out of scope

The integration test's TOCTOU in `preview_webhook_test.go`, and the suspected self-sustaining reconcile loop (no `GenerationChangedPredicate` on `For(&App{})`) — both left untouched.

### DIFF
--- a/internal/controller/app_controller.go
+++ b/internal/controller/app_controller.go
@@ -298,6 +298,10 @@ func (r *AppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	for _, env := range resolvedEnvs {
 		projectionEnvOrder = append(projectionEnvOrder, env.Name)
 	}
+	// Same selection updateStatus aggregates on. Computed once here so the env
+	// loop can refuse to write an app-global build failure it would only have
+	// to repair a moment later (CAI-229).
+	buildAggregationEnvNames := buildFailureAggregationEnvNames(resolvedEnvs, previewEnvNames)
 
 	for i := range resolvedEnvs {
 		env := &resolvedEnvs[i]
@@ -326,7 +330,7 @@ func (r *AppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		var image string
 		if app.Spec.Source.Type == mortisev1alpha1.SourceTypeGit && r.BuildClient != nil {
 			buildIdentity := resolveEnvBuildIdentity(&app, *env, previewBuildIdentities)
-			envImage, requeue, dirty, shouldClearNoCache, err := r.reconcileEnvBuild(ctx, &app, projectionEnvOrder, env.Name, buildIdentity.branch, buildIdentity.revision)
+			envImage, requeue, dirty, shouldClearNoCache, err := r.reconcileEnvBuild(ctx, &app, projectionEnvOrder, buildAggregationEnvNames, env.Name, buildIdentity.branch, buildIdentity.revision)
 			if err != nil {
 				return ctrl.Result{}, fmt.Errorf("reconcile buildrun for env %s: %w", env.Name, err)
 			}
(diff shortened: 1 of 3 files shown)


MC-Meesh:
Superseded by #510, which landed the same guard (buildFailureAggregationEnvNames threaded into reconcileEnvBuild) and closed CAI-229. Closing to keep one PR per fix.

